Issue #6, opened 2020-03-05
BERT embeddings are not contextualized


Arsilla:
Currently when using the 'sentence input with [token] input as this' put into the `SpacyLanguage` class does not result in the contextualized word embedding, since it will only input `token` into the model:

```
np.array_equal(bert['Going to the [store]'].vector, 
               bert['[store] this in the drawer please.'].vector)
>>> True
```
Spacy with Bert embeddings returns the context embedding using the token index:

```
np.array_equal(bertje('Going to the store')[3],
               bertje('store this in the drawer please.')[0])
>>> False
```

koaning:
Yep. This is a clear bug. Probably something wrong I am doing with the translating.

Arsilla:
AFAIK you need to input the whole sentence and use the index to get one token. The vector for the whole thing is then what would have been the CLS output in the original model (I’m guessing).

koaning:
Progress. 

```
from whatlies.language import SpacyLanguage

lang = SpacyLanguage("en_trf_robertabase_lg")
np.array_equal(
  lang.nlp("Going to the store")[3], 
  lang.nlp("Store this in the drawer")[0]
) # False 
```

The error is in my code.

koaning:
whoa. yep. that was totally a bug on my side. fixed now tho. 

<img width="832" alt="image" src="https://user-images.githubusercontent.com/1019791/76012351-6ebcaf00-5f16-11ea-924c-df29814f4621.png">

PR #11, opened 2020-03-10
Adding plot_graph_layout

Arsilla:
Function added to common.py and to embeddingset.

I also added an option of choosing the distance metric you want to use. I couldn't find any other layouts that allowed for a distance input.

### DIFF
--- a/whatlies/common.py
+++ b/whatlies/common.py
@@ -1,7 +1,15 @@
 import matplotlib.pylab as plt
+import networkx as nx
+
+import numpy as np
+import pandas as pd
+
+from sklearn.metrics.pairwise import distance_metrics
 
 
-def handle_2d_plot(embedding, kind, color=None, xlabel=None, ylabel=None, show_operations=False):
+
+def handle_2d_plot(embedding, kind, color=None, xlabel=None, ylabel=None, show_operations=False,
+                   annot=False):
     """
     Handles the logic to perform a 2d plot in matplotlib.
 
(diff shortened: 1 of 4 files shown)


whatlies/common.py
```diff
@@ -24,7 +32,32 @@ def handle_2d_plot(embedding, kind, color=None, xlabel=None, ylabel=None, show_o
         plt.quiver([0], [0], [embedding.vector[0]], [embedding.vector[1]], color=color,
                    angles='xy', scale_units='xy', scale=1)
         plt.text(embedding.vector[0] + 0.01, embedding.vector[1], name)
-    if kind == "text":
+    if (kind == "text") or annot:
```
koaning: i like this touch! but i'll likely change the term to `annotate` later. 

value(clarity) > value(saving some characters)

whatlies/language.py
```diff
@@ -28,16 +28,20 @@ def _input_str_legal(string):
...
-        if start != 0:
-            if end != -1:
-                vec = doc[start:end].vector
-        return Embedding(string, vec)
+        if end == -1:
+            end = start + 1
+        if bert:
+            new_str = string.replace("[", "").replace("]", "")
+            return Embedding(string, self.nlp(new_str)[start].vector)
+        else:
```
koaning: technically you don't need the else here.

whatlies/language.py
```diff
@@ -28,16 +28,20 @@ def _input_str_legal(string):
...
     def __getitem__(self, string):
         self._input_str_legal(string)
-        doc = self.nlp(string)
-        vec = doc.vector
+        bert = False
         start, end = 0, -1
         split_string = string.split(" ")
         for idx, word in enumerate(split_string):
             if word[0] == "[":
+                bert = True
```
koaning: i wonder why these changes were made. could you clarify?

koaning: technically ... `bert` is but a single type of contextualized embedding ... odds are there will be many other models too.

whatlies/language.py
```diff
@@ -28,16 +28,20 @@ def _input_str_legal(string):
...
                 end = idx + 1
-        if start != 0:
-            if end != -1:
-                vec = doc[start:end].vector
-        return Embedding(string, vec)
+        if end == -1:
+            end = start + 1
+        if bert:
+            new_str = string.replace("[", "").replace("]", "")
+            return Embedding(string, self.nlp(new_str)[start].vector)
```
koaning: i wonder why you only look at `start` here. 

these two examples should yield a different result; 

```
i am going to a [nice beach]
i am going to a [nice] beach
```

koaning:
@Arsilla thanks for the PR! I have some questions about the changes you've made to `__get_item__`. The rest looks green to merge.

whatlies/common.py
koaning: You don't have to change this for the PR if you don't feel like it. It is a minor details for this phase of the project.

whatlies/language.py
Arsilla: wow ok super weird, I made these local changes to fix the bugs from the issue #6. I swore I overwrote those with the fixes you've since made. But I guess something went from in my git history. 
I'll change change it back and push the changes tonight. 
(same goes for the other questions related to language.py)

whatlies/common.py
Arsilla: I guess I was inspired by the seaborn heatmap (I've used that a lot lately), they use annot :)

whatlies/language.py
koaning: i figured it might be something like that.

koaning:
LGTM

koaning:
thanks for the PR @Arsilla !

PR #32, opened 2020-04-24
fasttext support

koaning:


koaning:
We should also add a seperate language model. The fasttext API allows for dimensionality reduction on their side. 

image

PR #33, opened 2020-04-28
Fasttext spacy

koaning:


koaning:
Added fasttext support both via spaCy *and* native. Also added some warnings and tests.

PR #60, opened 2020-06-23
Fix EmbeddingSet getitem method

mkaze:
Currently, when a `tuple` or a `list` is used for getting items from an `EmbeddingSet` instance, it would return all the embeddings. This PR fixes this issue, so that a new `EmbeddingSet` is returned **containing only the requested embeddings**.

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -263,9 +263,9 @@ class EmbeddingSet:
         emb["buz"]
         ```
         """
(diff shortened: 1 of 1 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @tabergma will take a look at it as soon as possible ✨

koaning:
LGTM! Well spotted!

PR #66, opened 2020-06-23
Fix error in transformers test file

mkaze:
Since `nlp.vocab` or `vocab` in spaCy is just a cache of used tokens, therefore when the vocab or language is loaded from disk the `vocab` would be empty and therefore iterating over it would result in an empty list (as a result the `words` list would be empty as well). That's why currently 11 transformers tests fail. Instead, we need to use `vocab.strings` and iterate over that.

Also keep in mind that when you persist a `Vocab` or `Language` instance in spaCy, another special string denoted as `'_SP'` is automatically added to the list of strings in the vocab (see [here][1] or [here][2]).

[1]: https://github.com/RasaHQ/whatlies/blob/2783be06fcc2ccea01fd17da9cabfc7f3aacb987/tests/custom_test_vocab/strings.json#L29

[2]: https://github.com/RasaHQ/whatlies/blob/2783be06fcc2ccea01fd17da9cabfc7f3aacb987/tests/custom_test_lang/vocab/strings.json#L29

### DIFF
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -6,7 +6,7 @@ from whatlies.transformers import Umap, Pca, Noise, AddRandom
 
 
 vocab = Vocab().from_disk("tests/custom_test_vocab/")
-words = [v.text for v in vocab]
+words = list(vocab.strings)
 lang = SpacyLanguage(nlp=Language(vocab=vocab, meta={"lang": "en"}))
 emb = lang[words]
 


sara-tagger:
Thanks for submitting a pull request 🚀 @tmbo will take a look at it as soon as possible ✨

koaning:
LGTM.

Out of curiosity, would you know what `_SP` stands for?

mkaze:
> Out of curiosity, would you know what `_SP` stands for?

Actually, it a token representing *space* in part-of-speech tagger in spaCy for most of the languages (e.g. open [this link](https://github.com/explosion/spaCy/blob/master/website/docs/api/annotation.md) and search for `_SP`). And it seems spaCy clearly relies on it and manage it intentionally (e.g. see [here][1] or [here][2] or [here][3]). Though, I don't know the reason specially in the case when there is no tagger module present in vocab or language.

[1]: https://github.com/explosion/spaCy/blob/master/spacy/morphology.pyx#L153-L162
[2]: https://github.com/explosion/spaCy/blob/master/spacy/tests/serialize/test_serialize_vocab_strings.py#L41-L42
[3]: https://github.com/explosion/spaCy/blob/master/spacy/tests/serialize/test_serialize_vocab_strings.py#L56-L66

PR #67, opened 2020-06-23
Fix #62, Fix #64: makefile corrections

### LINKED ISSUE #62: "make develop" fails due to model mismatch
Currently, running `make develop` would install the small English model of spaCy (i.e. `en_core_web_sm`); however, the `tests/prepare_disk_for_tests.py` uses the medium version (i.e. `en_core_web_md`). This conflict results in the `make develop` command to fail in a clean development installation. Which one is intended to be used in the tests, the small or the medium version? It's also worth mentioning that in the notebooks only the small model have been used.

If you would like, I can push the fix as a PR.

### LINKED ISSUE #64: Using "setup.py develop" in makefile
I was wondering why the `python setup.py develop` has been used in `develop` target of makefile, especially when editable installation mode of `pip` have been used as well (i.e. `pip install -e` and `pip install -e .[dev]`). As far as I know, this not only makes it redundant, but also using `setup.py` is not recommended due to its potential conflicting behavior. Maybe it's there for good reasons which I am not aware of and would like to know.

If this is indeed a mistake, I can push a fix as a PR if you would like.

mkaze:
As the title says and discussed in related issues (#62, #64).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,8 @@ install:
 develop: install
 	python -m pip install -e ".[dev]"
 	python -m pip install --upgrade pip wheel setuptools twine
-	python setup.py develop
 	pre-commit install
-	python -m spacy download en_core_web_sm
+	python -m spacy download en_core_web_md
 	python -m spacy download en_trf_robertabase_lg
 	python tests/prepare_disk_for_tests.py
 	python tests/prepare_fasttext_tests.py


sara-tagger:

koaning:
LGTM. Thanks!

PR #68, opened 2020-06-23
Fix #63: Remove files option of black pre-commit hook

### LINKED ISSUE #63: "files" option in pre-commit config for black hook
A pattern, i.e. `files: sklego`, has been set for the filenames to be checked by `black` hook. Essentially, this makes `black` to skip all the files because nothing would match with that pattern. I was wondering if this is intentional or not.

If this is a mistake, I can work on a PR to fix it as well as run a check on all the files; however, let me know if you want to only the package files to be checked, i.e. set  `files: whatlies`, or include the tests as well, i.e. no need to set `files` option in that case.

mkaze:
This PR removes the `files` option of `black` hook so that all the code files, including tests, would be checked by `black`.

The repo needs to be checked by `black` since none of the files have been checked so far; but I think it's better to put that in another PR (if needed).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,4 +9,3 @@ repos:
     rev: stable
     hooks:
     -   id: black
-        files: sklego


sara-tagger:

koaning:
This is a fix for this issue: https://github.com/RasaHQ/whatlies/issues/63. Thanks!

Issue #35, opened 2020-04-28
add IVIS transformer


koaning:
https://bering-ivis.readthedocs.io/en/latest/python_package.html

sprintyaf:
Hi @koaning ,
I can work on adding Ivis transformer

PR #81, opened 2020-07-14
Ivis Transformer Implemented

sprintyaf:
As requested in the #35 issue, I added Ivis transformer. 
There are two arguments (k and batch_size) which depend on the size of the data. That's why I decided to fill them separately.
If there are any complaints, I'm ready to modify the code.


sprintyaf:
@koaning ,
I believe all errors are from another files. I can fix them if you want.
error

sara-tagger:
Thanks for submitting a pull request 🚀 @tttthomasssss will take a look at it as soon as possible ✨

koaning:
Mhmm ... that's strange. Feel free to fix, but more-over: we need to add some tests.

koaning:
Another thing: the `ivis` dependency needs to be added to the `setup.py` file as a dependency.

whatlies/transformers/ivis.py
```diff
@@ -0,0 +1,79 @@
...
+        self.kwargs = kwargs
+
+    def __call__(self, embset):
+        if not self.is_fitted:
+            self.fill_arguments(embset)
+            self.tfm = IVIS(embedding_dims=self.n_components, **self.kwargs)
+            self.fit(embset)
+        return self.transform(embset)
+
+    def fill_arguments(self, embset):
```
koaning, line 48: This can be a hidden method, the end user won't need to see it and I'd prefer it not show up in the documentation.

koaning: Come to think of it, do we really need this method? Why not just pass the `**kwargs` directly?

sprintyaf:
> Mhmm ... that's strange. Have you been forking from a recent master?
> 
> More-over: we need to add some tests. It'd be preferable if you could copy some of the tests from other transformers to also work for this one.
> 
> I recall having some issues in real life with this transformer in terms of installation, so I'd be good to confirm this.

About style errors, this file really has some long lines (line 11 and 19)
lines

I tried to run tests but I have some problems. Am I doing it right?
tests

sprintyaf:
> There's a discussion to be had on `fill_arguments`.

I get what you're saying. I just thought that if the size of the `embset` would be less than default values of `k` or `batch_size` (128 and 150), users would have to specify these values themselves. But it's okay I guess.

sprintyaf:
> > Mhmm ... that's strange. Have you been forking from a recent master?
> > More-over: we need to add some tests. It'd be preferable if you could copy some of the tests from other transformers to also work for this one.
> > I recall having some issues in real life with this transformer in terms of installation, so I'd be good to confirm this.
> 
> About style errors, this file really has some long lines (line 11 and 19)
> lines
> 
> I tried to run tests but I have some problems. Am I doing it right?
> tests

Oh I get it now, I was doing it wrong

setup.py
```diff
@@ -14,6 +14,7 @@
     "fasttext>=0.9.1",
     "bpemb>=0.3.0",
     "gensim>=3.8.3",
+    "ivis[cpu]",
```
koaning, line 17: Adding a version number would be preferable. Minimum number is fine.

koaning:
If the tests are having a hard time locally: have you run `make develop` locally? You can check the actions in the `Makefile`. The idea is that we prepare the tests such that we don't have to download Gigabytes of embeddings.

whatlies/transformers/ivis.py
```diff
@@ -0,0 +1,59 @@
...
+class Ivis:
+    """
+    This transformer scales all the vectors in an [EmbeddingSet][whatlies.embeddingset.EmbeddingSet]
+    by means of Ivis algorithm. We're using the implementation found in
+    [beringresearch/ivis](https://github.com/beringresearch/ivis)
+
+    Arguments:
+        n_components: the number of compoments to create/add
+        kwargs: keyword arguments passed to the Ivis implementation
+
```
koaning, line 17: Might be nice to link directly to the page that explains the hyperparams [here](https://bering-ivis.readthedocs.io/en/latest/hyperparameters.html).

tests/test_transformers.py
```diff
@@ -44,6 +46,7 @@ def test_transformations_new_size(transformer, extra_size):
         Noise(0.1),
         Tsne(2, n_iter=250),
         OpenTsne(2, n_iter=100),
+        Ivis(2, k=10, batch_size=10),
```
koaning, line 49: If you're listing hyperparams here that you consider important, it might be valid to upgrade them from `**kwargs` to normal keyword arguments.

koaning:
@sprintyaf it seems like you're not running black locally. This is probably due to the fact that you didn't start the pre-commit hooks. 

If you're unfamiliar with either of them, you may appreciate these tutorials on my blog on [pre-commit hooks](https://calmcode.io/pre-commit/the-problem.html) and [the black formatting tool](https://calmcode.io/black/introduction.html).

sprintyaf:
> @sprintyaf it seems like you're not running black locally. This is probably due to the fact that you didn't start the pre-commit hooks.
> 
> If you're unfamiliar with either of them, you may appreciate these tutorials on my blog on [pre-commit hooks](https://calmcode.io/pre-commit/the-problem.html) and [the black formatting tool](https://calmcode.io/black/introduction.html).

will do! thanks for the tutorials

sprintyaf:
due to some problems, I created a new pull request #106 
I'll close this one.

PR #85, opened 2020-07-18
Fix bug in movement_df method

mkaze:
To get the overlap, `intersection` should be used instead of `union`.


koaning:
Nice catch!

PR #94, opened 2020-07-23
Fix typos; improve wording

mkaze:
As the title says.


mkaze:
The failed test is probably a flaky test, since the vectors are generated randomly and therefore it is possible that vectors other than "dog" might be closer to "cat" vector. Note that adding a large number to "cat" and "dog" vectors does not necessarily make them similar in terms of cosine similarity (though, that trick may work if another metric, e.g. euclidean, is used instead).

sara-tagger:
Thanks for submitting a pull request 🚀 @tmbo will take a look at it as soon as possible ✨

PR #99, opened 2020-07-28
Add support for different ConveRT models and signatures

mkaze:
This PR improves `ConveRTLanguage` class implementation by adding the support for different ConveRT models and signatures. It also includes the relevant tests to ensure it's working properly for different configurations (although, this comes at the cost of longer running time of tests since the models should be downloaded and are rather large).

### DIFF
--- a/tests/test_lang/test_convert_lang.py
+++ b/tests/test_lang/test_convert_lang.py
@@ -4,12 +4,60 @@ from whatlies.language import ConveRTLanguage
 
 
 @pytest.fixture
-def lang():
-    lang = ConveRTLanguage()
+def lang(request):
+    lang = ConveRTLanguage(**request.param)
     return lang
 
 
-def test_basic_docs_usage1(lang):
+@pytest.mark.parametrize(
+    "lang, expected_shape",
+    [
+        ({"model_id": "convert", "signature": "default"}, (1024,)),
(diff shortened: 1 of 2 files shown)


koaning:
From glancing at the code this looks great! But .. I'm on holiday this week so I'll have a good look next monday. Thanks for the PR though!

mkaze:
@koaning I am not sure what to do for test models of this one. There is no minimal ConveRT model as far as I know.

koaning:
I'll merge it for now. It seems like things are running fast enough (5-6m is definately acceptable). We can worry about caching later. I also want to do some work to make some of the dependencies optional. But that is for another release.

koaning:
@mkaze by the way, got a linkedin/twitter account? You've sent a lot of meaningful PRs and this deserves a shout-out.

mkaze:
> by the way, got a linkedin/twitter account? You've sent a lot of meaningful PRs and this deserves a shout-out.

Thanks for your kind; I think those contributions were not significant, though. Anyways, I don't have a linkedin or twitter account. But if you happen to have an annoying unanswered TF/Keras/ML question on Stack Overflow, [you can reach me](https://stackoverflow.com/users/2099607/today?tab=profile) and (hopefully) there is a chance that I might be able to help you :)

Issue #40, opened 2020-05-04
Add tfhub support


koaning:
I imagine it would be nice to also have support for these: 

https://tfhub.dev/s?module-type=text-embedding&tf-version=tf2

koaning:
Good grief that stuff is instable. You really need to hard code versions in order for them to work on macos. Will delay this feature until later.

mkaze:
@koaning I think this one could be closed as well.

PR #100, opened 2020-07-28
Add support for TF Hub text-embedding models

mkaze:
This PR addresses issue #40. It provides (an initial) support for text-embedding models in TensorFlow Hub (or any other compatible model which could be loaded using `hub.load()`). Note that this only supports models which can operate/be called directly on raw text/sentence string; therefore, for the time being, it does not support models like BERT ([example](https://tfhub.dev/tensorflow/bert_en_uncased_L-12_H-768_A-12/2)) which need additional pre-processing steps.

### DIFF
--- a/tests/test_lang/test_tfhub_lang.py
+++ b/tests/test_lang/test_tfhub_lang.py
@@ -1,0 +1,33 @@
+import pytest
+
+from whatlies.language import TFHubLanguage
+
+
+@pytest.fixture
+def lang(request):
+    return TFHubLanguage(**request.param)
+
+
+@pytest.mark.parametrize(
+    "lang, expected_shape",
+    [
+        (
+            {"url": "https://tfhub.dev/google/universal-sentence-encoder-large/5"},
+            (512,),
+        ),
+        ({"url": "https://tfhub.dev/google/nnlm-en-dim50/2"}, (50,)),
+        (
+            {
+                "url": "https://tfhub.dev/google/tf2-preview/gnews-swivel-20dim-with-oov/1"
+            },
+            (20,),
+        ),
+    ],
+    indirect=["lang"],
+)
+def test_basic_usage(lang, expected_shape):
+    emb = lang["a test sentence"]
+    assert emb.vector.shape == expected_shape
+    emb = lang[["test", "a simple test sentence", "and another nice sentence"]]
+    assert len(emb) == 3
+    assert emb["test"].vector.shape == expected_shape
--- a/tests/test_lang/test_tfhub_lang.py
+++ b/tests/test_lang/test_tfhub_lang.py
@@ -11,11 +11,7 @@ def lang(request):
 @pytest.mark.parametrize(
     "lang, expected_shape",
     [
-        (
-            {"url": "https://tfhub.dev/google/universal-sentence-encoder-large/5"},
-            (512,),
-        ),
-        ({"url": "https://tfhub.dev/google/nnlm-en-dim50/2"}, (50,)),
+        ({"url": "https://tfhub.dev/google/tf2-preview/gnews-swivel-20dim/1"}, (20,),),
         (
             {
                 "url": "https://tfhub.dev/google/tf2-preview/gnews-swivel-20dim-with-oov/1"
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,6 @@
 [flake8]
 per-file-ignores =
   whatlies/language/convert_lang.py: F401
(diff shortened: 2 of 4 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @erohmensing will take a look at it as soon as possible ✨

erohmensing:
@koaning for you 🙂

tests/test_lang/test_tfhub_lang.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.fixture
+def lang(request):
+    return TFHubLanguage(**request.param)
+
+
+@pytest.mark.parametrize(
+    "lang, expected_shape",
+    [
+        (
+            {"url": "https://tfhub.dev/google/universal-sentence-encoder-large/5"},
```
koaning, line 15: I like the changes, but I'm unsure if this is an appropriate method of testing. Every time somebody might push a branch we'll download 500+ MB payload per model. 

Instead, is there a super small model that we might download? I'm also wondering if it makes sense to rely on `tf-hub` during testing. If the service ever goes down temporarily it would mess up our tests. Can we download a small model ahead of time instead? That's what I'm doing for `FastText` and `spaCy`.

koaning: I should acknowledge here that my initial `ConveRTLanguage` also suffers from this phenomenon.

mkaze: > Every time somebody might push a branch we'll download a 500+ MB payload per model.

There are two different aspects to this:
 1. It might be time-consuming to download a large file, so running tests would take longer: one workaround for the this issue for local dev and testing is to mark these tests as slow in pytest and then create a separate command which only runs non-slow tests. In this case, the developer can decide which one to run based on their modifications. Although, this does not work for Github actions, especially if it charges per minute.
 2. It might be a data transfer/storage limit in Github actions, or even on the developers side in their local dev environment: I don't have a solution for this issue!

> I'm unsure if this is an appropriate method of testing

I think some repos like [keras-applications](https://github.com/keras-team/keras-applications) (which basically consists of more than 10 models) use this approach. A single round of test takes around 10 minutes on Travis CI for keras-applications repo.

> is there a super small model that we might download?

Yes, there is. The swivel-20dim model which I have included in the tests is less than 2 MB. If you think we should not use large model files, we can only use that model instead.

> I'm also wondering if it makes sense to rely on tf-hub during testing...

If you ask me, I think it does make sense. Mainly because it's very rare that TF-HUB would be down (after all, it's maintained by Google!), and also I think the users would mostly rely on TF-HUB directly to use this class; so it makes sense to use TF-HUB directly in our tests. Although, even here we can instead download a small model (e.g. swivel-20dim) and put it in the test directory (however, I am not in favor of this approach entirely, even for fasttext or spacy model files as you have done before. That's because the dependencies are also changing and their format of saving/loading might change so those files go stale, which is the case for the current spacy model files in repo; and besides that even if we want to build/save/load a **minimal model** we should do it entirely in the tests to make sure that the library works with the most stable current version of dependencies).

**All in all, let me know of your final opinion on this and I'll make changes accordingly.**

koaning: Nice! I think your `swivel-20dim` suggestion is a good option. It'd allow us to fully test `tf-hub` as a backend which also means that we can do much lighter testing for items that depend on it like `ConveRT` and `universal-sentence-encoder`. 

*Anekdote* 

TF-HUB shouldn't go down often, that's true ... but I've had this one day where `s3` went down on amazon and because our CI depended on a project that had data hosted there (I think it was keras) our CI also broke because of it. Despite another cloud provider going down, we suddenly couldn't deploy anymore.

koaning:
Could you merge the changes from master into this? I just want to make sure this also works on the new tensorflow versions.

tests/test_lang/test_tfhub_lang.py
mkaze: All right, I just removed the large model files from the tests and used the two small variants of `swivel` model instead. Each has a size of ~1.5 MB.

> Anekdote...

I see! Thanks for sharing your experience.

mkaze:
@koaning Don't worry since I added the last commit after you merged the new dependency upgrade PR. So the current checks have used TF 2.3.

mkaze:
@koaning The failing test is the flaky test which I have mentioned earlier. Please re-run the checks. I will fix that test in another PR.

koaning:
LGTM. Thanks!

Issue #12, opened 2020-03-17
huggingface language model


koaning:
unless it can all be imported via spaCy

mkaze:
@koaning I think this could be closed as well.

PR #102, opened 2020-08-01
Add Hugging Face Transformers language class

mkaze:
This PR addresses #12 and implements `HFTransformersLanguage` class to support Hugging Face Transformer models. It's an initial implementation, with room for adding more features and custom configs. Currently, the embedding vector is obtained by summing the token representations (excluding special tokens, e.g. [CLS] or [SEP]).

Further, we have added `torch` as a **development dependency** so that we can test this class on both TensorFlow as well as PyTorch.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ base_packages = [
     "tensorflow==2.1.0",
     "tensorflow-text==2.1.1",
     "tensorflow-hub>=0.8.0",
+    "transformers>=3.0.0",
 ]
 
 docs_packages = [
@@ -35,6 +36,7 @@ test_packages = [
     "pytest-cov>=2.6.1",
     "nbval>=0.9.5",
     "pre-commit>=2.2.0",
+    "torch>=1.0.0",
 ]
 
 dev_packages = docs_packages + test_packages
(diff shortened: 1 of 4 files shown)


mkaze:
The checks should be re-run to see if the error in installing the package is resolved. If not, probably the hash in the cache is outdated and should be ignored with `--no-cache-dir` flag.

sara-tagger:
Thanks for submitting a pull request 🚀 @degiz will take a look at it as soon as possible ✨

mkaze:
@koaning As you may want, I removed large models from the tests of this class as well and replaced them with smaller models (~2MB). However, the error in installing torch is still there in Github actions.

koaning:
I've reset the job three times now. On the third time it passed. Dunno what caused it. Is this common?

mkaze:
@koaning No, I haven't seen such thing before (though, I don't use Github Actions a lot). Two times, it broke on installing torch due to hash conflict of downloaded wheel file. I don't why?!

logs

koaning:
My collegues are suggesting that github is having issues today that might also be causing some of this. I'll check again tomorrow morning. If it looks green, I'll merge.

mkaze:
@koaning All right, but I think the last run completed successfully with no errors (it's showing all green to me).

koaning:
Just re-ran everything and it ran fine. Cool.

Issue #91, opened 2020-07-23
spacy language backend sometimes produces wrong embeddings


mkaze:
The problem is that the embeddings given by the `SpacyLanguage` backend may not be correct in all cases. That's because currently a whitespace tokenizer [is used](1) for finding the position of tokens, whereas each language model in spaCy has its own tokenizer which might be more sophisticated than just a simple whitespace splitting. To demonstrate this, consider the following examples (note that the BERT model from `spacy-transformers` have been used here because the first two examples concern contextualized embeddings; however, it does not make a difference to use a built-in spaCy model instead):

```python
import numpy as np
from whatlies.language import SpacyLanguage

lang = SpacyLanguage("en_trf_bertbaseuncased_lg")

# =========================================
# Example 1
query_raw = "let's go [home]"
query_clean = "let's go home"
emb = lang[query_raw]
doc = lang.model(query_clean)
print(list(doc))
# [let, 's, go, home]

assert str(doc[2]) == 'go'
assert str(doc[3]) == 'home'
# In correct case, the following two statements should raise exception
assert np.array_equal(emb.vector, doc[2].vector)
assert not np.array_equal(emb.vector, doc[3].vector)

# =========================================
# Example 2
query_raw = "player #10 [won]"
query_clean = "player #10 won"
emb = lang[query_raw]
doc = lang.model(query_clean)
print(list(doc))
# [player, #, 10, won]

assert str(doc[2]) == '10'
assert str(doc[3]) == 'won'
# In correct case, the following two statements should raise exception
assert np.array_equal(emb.vector, doc[2].vector)
assert not np.array_equal(emb.vector, doc[3].vector)

# =========================================
# Example 3: even when there is no context emb
query_raw = "pre-order this product"
emb = lang[query_raw]
doc = lang.model(query_raw)
print(list(doc))
# [pre, -, order, this, product]

assert str(doc[:-2]) == "pre-order"
# In correct case, the following two statements should raise exception
assert np.array_equal(emb.vector, doc[:-2].vector)
assert not np.array_equal(emb.vector, doc.vector)
```
Note that we used an English model above, and spaCy supports multiple languages which might have complex tokenization rules beyond whitespace tokenization and we may not be aware of them (although we can look them up in the source code but that's not the point); so, we cannot just blame these non-alphanumeric characters (e.g. `'`, `#`, `-`, `?`, `!`) in English for the cause of this issue and treat these examples as special cases.

[...]

PR #103, opened 2020-08-02
Fix incorrect embedding bug in spaCy language class

mkaze:
This PR is a (temporary) fix for issue #91 which resulted from doing tokenization based on whitespace. I thought this was important to be fixed (until we find a better solution) because, as discussed in #91, even for some sentences (and languages with no space, such as Japanese) where there is no requested context embedding, the result is wrong as well.

@koaning I have used `en_core_web_md` model in the tests so that we could have better and more reliable real use-case tests. However, I did not add the installation command to Github actions config file yet (that's why the tests are failing) because I don't know if you are agreed with this or not. Let me know if you are OK with this, and I'll add another commit to this PR for that.

### DIFF
--- a/tests/test_lang/test_spacy_language.py
+++ b/tests/test_lang/test_spacy_language.py
@@ -4,7 +4,59 @@ import numpy as np
 from spacy.vocab import Vocab
 from spacy.language import Language
 from whatlies.language import SpacyLanguage
-from whatlies.language.spacy_lang import _selected_idx_spacy
+
+
+@pytest.fixture(scope="module")
+def lang():
+    return SpacyLanguage("en_core_web_md")
+
+
+def test_basic_usage(lang):
+    queries = [
+        "yesterday is history",
+        "tomorrow is mystery",
+        "today is a gift",
+        "that's why it's called [present]",
+    ]
+    emb = lang[queries]
+    assert len(emb) == 4
+    assert emb[queries[0]].name == "yesterday is history"
+    assert emb[queries[0]].vector.shape == (300,)
+
+
+@pytest.mark.parametrize(
+    "query",
+    ["no [closing bracket", "no opening] bracket", "[more] than one [context]",],
+)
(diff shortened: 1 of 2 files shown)


sara-tagger:

koaning:
The downside of using `en_core_web_md` is that it indeed is a pretty big payload and it feels like a bad practice to download large payloads on every commit. 

One thing I wonder ... can't you re-use the fixture for some of your tests? 

```python
@pytest.fixture()
def color_lang():
    vector_data = {
        "red": np.array([1.0, 0.0]),
        "green": np.array([0.5, 0.5]),
        "blue": np.array([0.0, 1.0]),
        "purple": np.array([0.0, 1.0]),
    }

    vocab = Vocab(strings=vector_data.keys())
    for word, vector in vector_data.items():
        vocab.set_vector(word, vector)
    nlp = Language(vocab=vocab)
    return SpacyLanguage(nlp)
```

Any unknown token should return a vector of zeros either way but all of your bracket-tests should still apply.

mkaze:
@koaning Yeah, I thought about that before adding new tests (considering that zero vector is return for OOV tokens); however, as I mentioned earlier, I thought it makes more sense to use a more real use-case for the tests. Anyways, if you think this is not a good idea, I'll make changes to the tests to use `color-lang` instead and submit it as a new commit to this PR.

koaning:
I'd prefer to have `color-lang` for now. It's more polite to the spaCy maintainers (I fear they pay for hosting the stuff) and it should also keep the tests nice and fast. Waiting for all these models to download is no fun in CI either. 

I'm open to ideas on how to handle this in the future though. We might train a more condensed spaCy model and add that a cached test.

mkaze:
@koaning All right, I just added a commit to use `color_lang` instead.

koaning:
It's looking good to me. I'll first perform an upgrade to a new version of tensorflow for the entire project. After that I'll merge this.

PR #106, opened 2020-08-03
Ivis Transformer Implementation

sprintyaf:
Hi, 
due to some problems, I decided to start over from a new branch. I installed whatlies with make develop, installed pre-commit hooks and ran black formatting on files that I've changed. All tests are passing locally.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ base_packages = [
     "tensorflow>=2.3.0",
     "tensorflow-text>=2.3.0",
     "tensorflow-hub>=0.8.0",
+    "ivis[cpu]>=1.8.0",
 ]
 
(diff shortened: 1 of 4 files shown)


sprintyaf:
@koaning,
about hyperparameters that you mentioned in the last PR, I only specified them because with their default values transformer was not working. The default values of "k" and "batch size" are 150 and 128, respectively. If the number of samples is greater than those values, Ivis transformer is not working and needs to set those values to be less than or equal to the number of samples.

sara-tagger:
Thanks for submitting a pull request 🚀 @akelad will take a look at it as soon as possible ✨

koaning:
Looks good to me! Thanks for the PR.

PR #108, opened 2020-08-04
Fix flaky test of embeddingset

mkaze:
This PR fixes the infamous(!) flaky test. Considering the embedding vectors in `lang`, instead of `cosine` we use `euclidean` as the distance metric to prevent similarity error due to happenstance.


koaning:
+1

PR #110, opened 2020-08-04
Upgrade setuptools and wheel in github actions workflow

mkaze:
As the title says.


koaning:
This might fix the torch install issues?

mkaze:
@koaning I don't think so (that was a hash mismatch issue). I did this because one or two dependencies complained about incompatible setuptools version in Github Actions, and also I think it's a good practice as well to have the latest version of them in the tests.

koaning:
That's fair. Thanks!

PR #111, opened 2020-08-05
Optional deps

koaning:


mkaze:
Note that `transformers` package works with either of TF or PyTorch (i.e. if one has TF, there is no need to install `torch`). So, `torch` is not a **strict** requirement of that package. However, I added `torch` to `test_packages` in `setup.py` to ensure in the tests that `HFTransformersLanguage` works with both of TF and PyTorch.

mkaze:
And one more suggestion: instead of splitting deps per language class (which I previously suggested), alternatively this could be done with just base, i.e. `pip install whatlies` (which may contain smaller packages like fasttext, spacy, gensim, etc.), and another one for all languages i.e. `pip install whatlies[full]` which includes the more involved ones like TF-Hub or HFTransformers as well. Though, this also has its own pros/cons compared to current approach.

koaning:
@mkaze I currently have  

```
pip install whatlies[all]
```

That should install it all. Would you argue that `full` is a more common term?

mkaze:
@koaning Sorry, I saw `ivis_dep` and `open_tsne_dep` and thought you have made a install command for each language class as well. But it seems that's not the case. So you can ignore my comment. As for the naming, both `full` and `all` seem good to me.

koaning:
My goal is to release 0.4.5 today. That way we can collect feedback before putting everything in 0.5.0. The changes you've added certainly warrant an upgrade from 0.4 to 0.5.

mkaze:
@koaning Why did you remove `torch` from `test_packages` as well? As I mentioned earlier, it's used in the tests!

There is a minor flaw in your recent change: we should definitely use `[dev]` for tests and not the `[all]` (e.g. `[all]` does not contain `torch` which is used in the tests; this would even fail if you add `torch` to `test_packages`). However, you should set them correctly:

```python
extra_deps = tf_packages + transformers_dep + ivis_dep + open_tsne_dep
dev_packages = docs_packages + test_packages + extra_deps

#...
 extras_require={
    #...
    'all': extra_deps,
}
```

This is because when running `pip install package[xyz]` all the base packages, i.e. those in `install_requires`, would be installed by default. So `xyz` should only contain the additional packages.

koaning:
@mkaze I've removed it temporarily to figure what why the hash install pip thing re-appeared. It'll be added before I git merge. I'm upgrading torch to at least `1.6.0`. Maybe that will help. 

Your second point is a good one. Will make the change :)

mkaze:
@koaning I see. I intentionally set that to `1.0.0` to cover a wider range of users with different/older versions (i.e. it does not upgrade the user's `torch` installation unnecessarily). But using `1.6.0` is also okay. If you remember I had the same opinion about TF.

This issue of dependency versions should also be dealt with at a later time (e.g. unfortunately, `pip` does not have a dependency resolver and therefore there are now some conflicts with `tensorflow` and `numpy` or `scipy` versions which will be installed).

koaning:
Yeah pip has some downsides, but it still seems like the most widely accepted base-tool. This build is looking green now. 

Would you be cool if I released this as the new version?

mkaze:
> Would you be cool if I released this as the new version?

Sure, no problem :)

mkaze:
@koaning Oh, please! Don't do this to yourself and me :smile: You forgot to:

1. add `extra_deps` to `dev_packages`.
2. revert back `[all]` to `[dev]` in the test workflow.

mkaze:
@koaning This regular hash mismatch error for `torch` is so annoying. Three reasons comes to my mind for the cause of this error:

- The downloaded wheel file is not complete, i.e. gets corrupted during download (this becomes more likely considering that it's a large file +700 MB),
- The Github Actions has an underlying problem (e.g. network, etc.),
- The file on PyPI server or the server itself is flaky (although, this is probably the most unlikely reason because it would have been reported and fixed in the past days).

koaning:
Yeah ...  I'm having trouble figuring out what it is exactly. Will try out some versions.

koaning:
I fear pytorch is a beast to install on linux. 

image

image

mkaze:
@koaning I don't know! It's strange to me, though! That's because I always work in Ubuntu and have installed `torch` as dependency of different projects many times without encountering an issue like that. For example, packages like `spacy-transformers` (see [here](https://github.com/explosion/spacy-transformers/blob/1280c8b7236d8708ed2f82a4251bd1e18296dd37/requirements.txt#L3)) or even Captum (see [here](https://github.com/pytorch/captum/blob/e215a5af2629c7e2fb899dbcf417a727fc196f16/setup.py#L139)) which is an official PyTorch project just simply list `torch` in their requirements or `setup.py` without any additional options.

koaning:
I'm running the whole thing with an if statement in there checking for OS. We'll see if that works.

mkaze:
@koaning The problem is that previously it worked one time, and did not work another time. This makes it harder to debug!

BTW, there is no need to install `torchvision`. That's for CV and not needed here.

.github/workflows/pythonpackage.yml
```diff
@@ -15,19 +15,27 @@ jobs:
...
       run: |
         python -m pip install --upgrade pip setuptools wheel
         pip install -e ".[dev]"
         pip install flake8 pytest black
+        pip install torch>=1.6.0+cpu torchvision==0.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+    - name: Install dependencies Ubuntu
+      if: ${{matrix.os == 'macos-10.15'}}
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install -e ".[all]"
```
mkaze, line 37: Change this to `[dev]` :)

.github/workflows/pythonpackage.yml
```diff
@@ -15,19 +15,27 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install dependencies MacOS
+      if: ${{matrix.os == 'ubuntu-latest'}}
       run: |
         python -m pip install --upgrade pip setuptools wheel
         pip install -e ".[dev]"
         pip install flake8 pytest black
+        pip install torch>=1.6.0+cpu torchvision==0.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+    - name: Install dependencies Ubuntu
```
mkaze, line 33: Install dependencies MacOS

.github/workflows/pythonpackage.yml
```diff
@@ -15,19 +15,27 @@ jobs:
...
+        python-version: [3.7]
 
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install dependencies MacOS
```
mkaze, line 26: Install dependencies Ubuntu

.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,46 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install flake8 pytest black
+    - name: Run Base Install
+      run: |
+        python -m pip install -e ".[base]"
```
mkaze, line 26: `[base]` is redundant here.

.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,46 @@
...
+        python tests/check_pip.py contains spacy tensorflow
+    - name: Install IVIS
+      run: |
+        python -m pip install -e ".[ivis]"
+        python tests/check_pip.py missing opentsne
+        python tests/check_pip.py contains spacy tensorflow ivis
+    - name: Install All
+      run: |
+        python -m pip install -e ".[all]"
+        python tests/check_pip.py contains spacy tensorflow ivis opentsne ivis
```
mkaze, line 42: Change the second `ivis` to `transformers`.

.github/workflows/pythonpackage.yml
```diff
@@ -15,19 +15,27 @@ jobs:
...
+      if: ${{matrix.os == 'ubuntu-latest'}}
       run: |
         python -m pip install --upgrade pip setuptools wheel
         pip install -e ".[dev]"
         pip install flake8 pytest black
+        pip install torch>=1.6.0+cpu torchvision==0.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
mkaze, line 32: Remove `torchvision==0.7.0+cpu`.

.github/workflows/pythonpackage.yml
```diff
@@ -15,15 +15,23 @@ jobs:
...
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install -e ".[dev]"
+        pip install flake8 pytest black
+        pip install torch>=1.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+    - name: Install dependencies MacOS
+      if: ${{matrix.os == 'macos-10.15'}}
       run: |
         python -m pip install --upgrade pip setuptools wheel
         pip install -e ".[dev]"
```
mkaze, line 37: OK, the Ubuntu checks were successful. Now, you can add `pip install torch` before or after this line to install it for MacOS as well (this is needed because `torch` is removed entirely from `setup.py`).

.github/workflows/croncheck.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+    # Runs a single command using the runners shell
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
```
mkaze, line 26: The same approach should be done for this workflow as well.

koaning: Ah yeah. Forgot about that.

.github/workflows/pythonpackage.yml
koaning: d0h.

mkaze:
@koaning Finally! Great job! :+1:

mkaze:
@koaning And now we should be %100 sure that [Github Actions is to blame](https://github.com/actions/virtual-environments/issues/1354)!..... I know how you feel now after all that efforts; please accept my sincere sympathies on the loss of so much of your energy and time! :wink:

PR #115, opened 2020-08-06
Fix bug in score_similar method of EmbeddingSet

mkaze:
The following raises an error:

```python
foo = Embedding('foo',...)
bar = Embedding('bar',...)
baz = Embedding('baz',...)

embset = EmbeddingSet(foo, bar)
embset.score_similar(baz, n=2)

> ValueError: Embedding for `baz` does not exist in this EmbeddingSet
```

Only if the given embedding is a string it should be checked whether it exists in `EmbeddingSet` or not. Otherwise, an `Embedding` instance is always allowed.

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -403,12 +403,11 @@ class EmbeddingSet:
                 f"You cannot retreive (n={n}) more items than exist in the Embeddingset (len={len(self)})"
             )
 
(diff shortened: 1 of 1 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @TyDunn will take a look at it as soon as possible ✨

PR #126, opened 2020-08-12
[WIP] Cosine Distance

koaning:
This is a work in progress PR for now. I discovered some bugs while working on a an article. This PR hopes to adress it while also adding some more intuitive verbs.

### DIFF
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -177,7 +177,7 @@ class Embedding:
 
     def plot(
         self,
-        kind: str = "scatter",
+        kind: str = "arrow",
         x_axis: Union[str, "Embedding"] = None,
         y_axis: Union[str, "Embedding"] = None,
         color: str = None,
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -549,6 +549,42 @@ class EmbeddingSet:
         plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
         plt.show()
 
+    def plot_distance(self, metric=None):
+        """
+        Make a correlation plot. Shows you the correlation between all the word embeddings. Can
+        also be configured to show distances instead.
+
+        Arguments:
+            metric: don't plot correlation but a distance measure, must be scipy compatible (cosine, euclidean, etc)
+
+        Usage:
+
+        ```python
+        from whatlies.language import SpacyLanguage
+        lang = SpacyLanguage("en_core_web_md")
+
+        names = ['red', 'blue', 'green', 'yellow', 'cat', 'dog', 'mouse', 'rat', 'bike', 'car']
+        emb = lang[names]
+        emb.plot_distance()
+        ```
+
+        ![](https://rasahq.github.io/whatlies/images/corrplot.png)
+        """
+        df = self.to_dataframe().T
+        corr_df = (
+            pairwise_distances(self.to_matrix(), metric=metric) if metric else df.corr()
+        )
+
+        fig, ax = plt.subplots()
+        plt.imshow(corr_df, cmap=plt.cm.get_cmap().reversed())
+        plt.xticks(range(len(df.columns)), df.columns)
+        plt.yticks(range(len(df.columns)), df.columns)
+        plt.colorbar()
+
+        # Rotate the tick labels and set their alignment.
+        plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
+        plt.show()
+
     def plot_pixels(self):
         """
         Makes a pixelchart of every embedding in the set.
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -7,10 +7,14 @@ import pandas as pd
 import matplotlib.pylab as plt
 import altair as alt
 from sklearn.metrics import pairwise_distances
-from sklearn.metrics.pairwise import paired_distances
+from sklearn.metrics.pairwise import (
+    paired_distances,
+    cosine_similarity,
+    cosine_distances,
+)
 
 from whatlies.embedding import Embedding
-from whatlies.common import plot_graph_layout
+from whatlies.common import plot_graph_layout, normalize
 
 
 class EmbeddingSet:
@@ -181,7 +185,7 @@ class EmbeddingSet:
         if mapping == "direct":
             return [v > other for k, v in self.embeddings.items()]
 
-    def to_X(self):
+    def to_X(self, norm=False):
         """
         Takes every vector in each embedding and turns it into a scikit-learn compatible `X` matrix.
 
@@ -199,7 +203,12 @@ class EmbeddingSet:
         X = emb.to_X()
         ```
         """
-        X = np.array([i.vector for i in self.embeddings.values()])
+        X = np.array(
+            [
+                normalize(i.vector) if norm else i.vector
+                for i in self.embeddings.values()
+            ]
+        )
         return X
 
     def to_X_y(self, y_label):
@@ -549,13 +558,12 @@ class EmbeddingSet:
         plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
         plt.show()
 
-    def plot_distance(self, metric=None):
+    def plot_similarity(self, metric="cosine", norm=True):
         """
-        Make a correlation plot. Shows you the correlation between all the word embeddings. Can
-        also be configured to show distances instead.
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
 
         Arguments:
-            metric: don't plot correlation but a distance measure, must be scipy compatible (cosine, euclidean, etc)
+            metric: cosine or euclidean
 
         Usage:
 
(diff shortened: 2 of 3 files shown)


koaning:
This branch will also fix https://github.com/RasaHQ/whatlies/issues/130

whatlies/embeddingset.py
```diff
@@ -549,6 +568,81 @@ def plot_correlation(self, metric=None):
...
+        lang = SpacyLanguage("en_core_web_md")
+
+        names = ['red', 'blue', 'green', 'yellow', 'cat', 'dog', 'mouse', 'rat', 'bike', 'car']
+        emb = lang[names]
+        emb.plot_distance()
+        ```
+
+        ![](https://rasahq.github.io/whatlies/images/corrplot.png)
+        """
+        df = self.to_dataframe().T
```
mkaze, line 591: I was wondering what's the point of using `to_dataframe` here. As it seems, you are just using the `columns` attribute of this dataframe which is basically equivalent to `list(self.embeddings.keys())`!

whatlies/embeddingset.py
```diff
@@ -549,6 +568,81 @@ def plot_correlation(self, metric=None):
...
```
mkaze, line 629: Same as my above comment!

whatlies/embeddingset.py
```diff
@@ -199,7 +213,12 @@ def to_X(self):
         X = emb.to_X()
         ```
         """
-        X = np.array([i.vector for i in self.embeddings.values()])
+        X = np.array(
```
mkaze, line 216: Instead, you could have simply used `sklearn.preprocessing.normalize` and write:

```python
from sklearn.preprocessing import normalize

def to_X(self):
    X = np.array([i.vector for i in self.embeddings.values()])
    X = normalize(X) if norm else X
    return X
```

mkaze:
@koaning Stop the automated tests! I suspect there is something wrong with ivis tests, but first I need to make sure.

whatlies/embedding.py
```diff
@@ -177,7 +177,7 @@ def distance(self, other, metric: str = "cosine"):
 
     def plot(
         self,
-        kind: str = "scatter",
+        kind: str = "arrow",
```
mkaze, line 180: This might be a breaking change for some users, so at least they should be clearly notified of this in release notes!
As a side point, I think one of the standards of PRs is to make them as small and relevant as possible and concerned with only one specific feature/issue.

koaning:
MHm ... you're probably right on the IVIS test. I don't get why it fails only sometimes tho.

whatlies/embeddingset.py
koaning: It's an artifact left over from the previous implementation. Originally I also wanted to have `correlation` as a metric which is super easy to calculate via `df.corr()`. The code definately is work in progress. I certianly think a cleanup is possible here.

whatlies/embedding.py
koaning: You're right. This PR started as a "branch of things to fix for a paper we're writing". It's still a work in progress though. 

You're right about the release notes. I've been adding them to github but it might be time to start adding them to the documentation as well.

mkaze:
@koaning Yes, it's likely that the Ivis tests have an issue (the tests after OpenTsne are those of Ivis : [here][1] and [here][2]). I am not familiar with ivis and its method, but maybe it does not converge due to a bad random initialization?! Not sure!

[1]: https://github.com/RasaHQ/whatlies/pull/126/checks?check_run_id=976117223
[2]: https://github.com/RasaHQ/whatlies/runs/966126745?check_suite_focus=true

koaning:
I've made a new PR for IVIS tests here: https://github.com/RasaHQ/whatlies/pull/131.

koaning:
@mkaze I'll be releasing a video as part of my algorithm whiteboard series on debias-ing. Since I'm using this the features in this branch I will need to push a version live this week. To keep things basic ... would you be opposed to just supporting `cosine` and `euclidean` distance metrics for now for the `plot_distance` and `plot_similarity` charts?

mkaze:
> To keep things basic ... would you be opposed to just supporting 

@koaning No problem, it's fine to me.

By the way, the `norm` argument for `to_X` is missing.

koaning:
@mkaze There's still some silly things in the CI it seems. Super long install times for Mac. 

image

koaning:
Also really long waiting time for the tests. 

image

koaning:
This was an interesting one. The source code had `plt.show()` in it which was causing the MacOS tests to block. Was able to confirm this locally. Hopefully it is now fixed.

koaning:
@mkaze it's ready for review now.

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
...
+        plt.xticks(range(len(self)), self.embeddings.keys())
+        plt.yticks(range(len(self)), self.embeddings.keys())
+        plt.colorbar()
+
+        # Rotate the tick labels and set their alignment.
+        plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
+
+    def plot_distance(self, metric="cosine", norm=False):
+        """
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
```
mkaze, line 639: "similarity" should be replaced with "distance" (two instances).

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
         # Rotate the tick labels and set their alignment.
         plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
 
+    def plot_similarity(self, metric="cosine", norm=False):
+        """
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
+
+        Arguments:
+            metric: `cosine` or `correlation`
+            norm: normalise the embeddings before calculating the distance
```
mkaze, line 601: ...before calculating the "similarity", right?

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
         # Rotate the tick labels and set their alignment.
         plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
 
+    def plot_similarity(self, metric="cosine", norm=False):
+        """
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
+
+        Arguments:
+            metric: `cosine` or `correlation`
```
mkaze, line 600: I think these should be wrapped in single or double quotation to make it clear they are strings, i.e. `'cosine'` or `'correlation'`.

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
...
+
+        # Rotate the tick labels and set their alignment.
+        plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
+
+    def plot_distance(self, metric="cosine", norm=False):
+        """
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
+
+        Arguments:
+            metric: `cosine`, `correlation` or `euclidean`
```
mkaze, line 642: Same as the previous comment.

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
...
+            raise ValueError(
+                f"The `metric` argument must be in {allowed_metrics}, got: {metric}."
+            )
+
+        vmin, vmax = 0, 1
+        X = self.to_X(norm=norm)
+        if metric == "cosine":
+            distances = cosine_distances(X)
+        if metric == "correlation":
+            distances = 1 - np.corrcoef(self.to_X())
```
mkaze, line 669: The values of Pearson corr coefficients are between -1 and 1. So do you want to scale the values to [0, 1] considering that you have set `vmin, vmax = 0, 1`?

koaning: Well spotted!

whatlies/embeddingset.py
```diff
@@ -580,6 +592,94 @@ def plot_correlation(self, metric=None):
...
+            raise ValueError(
+                f"The `metric` argument must be in {allowed_metrics}, got: {metric}."
+            )
+
+        vmin, vmax = 0, 1
+        X = self.to_X(norm=norm)
+        if metric == "cosine":
+            similarity = cosine_similarity(X)
+        if metric == "correlation":
+            similarity = np.corrcoef(self.to_X())
```
mkaze: Same as the comment I made about correlation.

koaning:
Ah I just pushed because I thought you were done reviewing. My bad. Will wait for a bit now.

whatlies/embeddingset.py
```diff
@@ -580,6 +592,96 @@ def plot_correlation(self, metric=None):
         # Rotate the tick labels and set their alignment.
         plt.setp(ax.get_xticklabels(), rotation=90, ha="right", rotation_mode="anchor")
 
+    def plot_similarity(self, metric="cosine", norm=False):
+        """
+        Make a similarity plot. Shows you the similarity between all the word embeddings in the set.
+
+        Arguments:
+            metric: 'cosine' or 'correlation'
```
mkaze, line 600: Oh, sorry! I think I didn't explained it in full. The backticks (`) should remain there (so that they could be shown properly, i.e. as inline code snippet, in the documentation).

koaning: Ah! Good call.

mkaze:
@koaning Sorry that I make some annoying reviews sometimes :smile:

It looks good to me! Well done! :+1:

koaning:
> @koaning Sorry that I make some annoying reviews sometimes 😄

No worries. Better to be strict and picky than to allow everything.

koaning:
I'll make a new release this weekend with this plot in it. There's some videos on the algorithm whiteboard coming out on Monday that use this distance map.

PR #135, opened 2020-08-12
Extra Sklearn Test

koaning:
I was wondering if the scikit-learn implementation is robust against duplicates (which can totally happen). So I extended the test to confirm.


koaning:
It is looking green.

mkaze:
Yeah, a good thing to consider. Since the embeddings for the input data are generated one-by-one, instead of feeding them all to the language class to get an `EmbeddingSet` instance, the duplicates are preserved.

PR #136, opened 2020-08-14
Add FAQ on Altair Customization. 

koaning:


mkaze:
Looks good to me!
A sideline question just out of curiosity: I have seen you use semicolon instead of colon in your writings as well as the documentation; I was wondering if this is a common practice, since I have not encountered this before.

koaning:
I just looked it up. It seems that they do the same thing (indicate that a list is about to follow) but a semicolon needs to be used inline; like this, this or like this. While a colon is followed by a list, such as: 

- this 
- this 
- and this 

... I'll make the edit.

PR #137, opened 2020-08-14
Fix #124 - Allow one Embedding to be passed to EmbeddingSet constructure

### LINKED ISSUE #124: Allow embeddings to be passed as *args
If `foo` is an `Embedding` instance, `EmbeddingSet(foo)` is not supported. Currently we need to write;

```
EmbeddingSet({'foo': foo})
```

It would be nicer if we also allow the aforementioned approach.

mkaze:
Fixes #124, as the title says.

### DIFF
--- a/tests/test_embeddingset.py
+++ b/tests/test_embeddingset.py
@@ -24,6 +24,22 @@ def lang():
     return SpacyLanguage(nlp)
 
 
+def test_embeddingset_creation():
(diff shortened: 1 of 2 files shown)


koaning:
The change looks good, but maybe add a unit test?

mkaze:
The tests were added.

PR #141, opened 2020-08-14
Fix #139: Conditional Imports

### LINKED ISSUE #139: Dependency Installation goes Awry
image

If I have a new virtualenv with no `tensorflow-text` then we can still get silly import errors. Need to fix this.

koaning:
This is a work in progress PR for https://github.com/RasaHQ/whatlies/issues/139. I'll start by adding tests that capture what goes wrong and then I'll add a fix.


koaning:
Awesome. We've got our failing test which means we can proceed.

mkaze:
Two notes while you are working on this:

- Test for `[transformers]` and `[opentsne]` are missing.
- Just a suggestion: you may want to create a new test file which just do some imports and then test it using pytest in each stage:
    ```python
    from whatlies import *
    from whatlies.language import *
    from whatlies.transformers import *
    ```
    and then in the workflow file, you can use `pytest tests/test_imports.py` at then of each `run` stage.

koaning:
@mkaze I think I've got a nice little fix going here. We now show a pretty helpful error if a dependency is missing. 

image

koaning:
We currently have a lot of scripts in the `tests` folder. Things have accumulated. I'd love to clean it up but I think I'll do that in a seperate PR to keep it clear what this one is about.

mkaze:
@koaning The error message is very helpful; well done!

koaning:
Any comments @mkaze or shall I merge?

mkaze:
@koaning Looks good to me! Thanks a lot for this!

PR #142, opened 2020-08-14
moved into sep scripts folder

koaning:
This is an attempt to clean up the script files and to place them into a single folder.


mkaze:
Looks good to me. The make file also has [some references](https://github.com/RasaHQ/whatlies/blob/master/Makefile#L11-L13) that need to be changed.

mkaze:
Oh, sorry! I missed it, you have already done that :smile:

PR #144, opened 2020-08-14
Automated Tests on Documentation

koaning:
This is the first step to automate our docs a bit more. We're now parsing the docstrings and checking if they cause an error. 

Turns out ... they no longer do.


mkaze:
I think you can even safely remove the OS check for installing `torch` as `pip install torch` should work for both.

koaning:
I'll check that! I think for this PR it may be better to only add the tests for the apidoc. The tutorial docs contain really heavy language backends.

koaning:
Mhm. It's getting back to flaky behavior again. 

image

koaning:
I'll restart the workflow to make sure it's not a github actions hiccup. If the slowness persists I'll revert back to the original torch install to see if it makes a difference.

Issue #140, opened 2020-08-14
The return value of plot methods


mkaze:
*This is partly relevant to #127.*

 Currently, the following methods return a reference to the plot objects:
- `plot_interactive_matrix`
- `plot_interactive`
- `plot_movement`

However, the following methods either return `None` or `self`:
- `plot_pixels`
- `plot_correlation`
- `plot_graph_layout`
- `plot` from `EmbeddingSet`
- `plot` from `Embedding`

Isn't it better that the methods in the latter group also return a reference to the plot object? This makes them all have similar interface, as well as gives the user the ability to tune the plot object further if desired (although, the methods in the second group use `matplotlib` and therefore the plots could be customized with `plt.the_relevant_function(...)` as well).

koaning:
I certainly agree that we could apply some more consistency there. Sofar I've been able to use hacks like this in the notebook to suit my needs: 

```
plt.figure(figsize=(8,8))
emb.plot_pixels()
```

If we could maintain the ability to allow for these hacks in the notebook that'd be neat but I agree that proper consistency is more important.

koaning:
One caveat: the `plot` method from the `Embedding` is really only meant for demo purposes. I am also considering deprecating `plot_correlation` in favor of `plot_distance` and `plot_similarity`.

mkaze:
> the plot method from the Embedding is really only meant for demo purposes

If it's there and also there is no plan to deprecate it, then it should work properly, right? I am asking this because I have found one problem with it in a rather niche case: it does not project the 2D-embeddings if `x_axis` and/or `y_axis` are `Embedding` instances; rather, it just draw the embedding itself. Though, it will be probably fixed in the PR for #125.

mkaze:
Unfortunately, this is a breaking change especially that chained `plot` calls have been used [in the tutorials](https://rasahq.github.io/whatlies/tutorial/embeddings/#plotting-high-dimensions). @koaning I don't think you be in favor of it in that case.

koaning:
If I'm honest, I think the library doesn't have *that* many users yet. Here's the views for the relevant documentation page; 

<img width="1259" alt="image" src="https://user-images.githubusercontent.com/1019791/90272873-5ace8f00-de5e-11ea-90a4-dc74ab8fba9c.png">

[...]

Issue #127, opened 2020-08-12
Add support for customizing axis label as well as title of the plots


mkaze:
Currently, the user cannot customize the axis label and the title of a plot. It's good if we could provide a built-in support for this.

sara-tagger:
Thanks for the issue, @JustinaPetr will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

koaning:
With the advent of features like https://github.com/RasaHQ/whatlies/issues/129 I wonder what the best approach is here. One concern I have is that the plot method might get "super full of variables". Especially if we're also considering another backend like Bokeh. 

Might it be enough to document how to work with Altair? Something like; 

```python
(lang[text]
  .transform(Umap(2))
  .plot_interactive("umap_0", "umap_1", annot=False)
  .properties(title="spaCy", height=200, width=200))
```

If possible, I'd prefer to relay customisation of the plots to the user. The feature list of customisation options can get really big.

mkaze:
> If possible, I'd prefer to relay customisation of the plots to the user. The feature list of customisation options can get really big.

All right, agreed If you think that's a lot to offer and might not noticeably make the user life easier; although, personally, when I see a method with lots of arguments I cheer up because it gives more freedom to do various things in a builtin-way and without me to take care of it implementation. On the other hand, that's also a fair point to consider the current stage and scope of the library.

koaning:
For this issue then we'd consider an addition to the documentation to also be a valid solution?

mkaze:
> For this issue then we'd consider an addition to the documentation to also be a valid solution?

Yeah, if the plan is to not support these. We can add a note or an example to the documentation or FAQ section for further configuring the plots.

koaning:
I'll make a PR that adds this to the FAQ. That should help.

koaning:
I've made a PR here https://github.com/RasaHQ/whatlies/pull/136 but it seems like renaming the axes is not supported via `.properties`. This is making me wonder if we do need to add it as a feature. I can certainly see people wanting to customise the axes.

koaning:
I've stumbled across a related issue. 

<img width="1178" alt="image" src="https://user-images.githubusercontent.com/1019791/90243999-32c83700-de30-11ea-992b-ac9fb0e87a60.png">

[...]

Issue #125, opened 2020-08-12
Integer Default Parameters for Plotting Methods


koaning:
The goal is to allow for this syntax; 

```python
emb.transform(Pca(2)).plot(x_axis=0, y_axis=1)
```

The reasoning is currently you have to write;

```python
emb.transform(Pca(2)).plot(x_axis="pca_0", y_axis="pca_1")
```

This get's annoying when you want to replace the `Pca` with `Umap`. 

```python
emb.transform(Umap(2)).plot(x_axis="umap_0", y_axis="umap_1")
```

What's even nicer is that with proper default arguments in the future you can just run; 

```python
emb.transform(Umap(2)).plot()
```

mkaze:
I am working on a PR addressing this.

mkaze:
@koaning One question: what would be the appropriate default axis labels in this case?

koaning:
I think `x_axis=0` and `y_axis=1` should suffice, no?

mkaze:
No, I mean the values for `xlabel` and `ylabel` in `handle_2d_plot`. For example, "D0" or "Dim 0" is good?

koaning:
In the case that an `Embedding` is passed I would still much prefer that we use the associated name. I suppose if we use a dimension then `Dimension X` where `X` is the index?

mkaze:
@koaning Having implemented and merged this feature, now how do we want to go about handling those `pca_0`, `pca_1`, etc. embeddings? Do we want to entirely remove them, or make them optional (e.g. for backwards compatibility) by adding a separate argument in the `__init__` method of corresponding transformer classes, e.g. `def __init__(self, ..., add_indicators=False)`? Or any other option?

koaning:
Something inside me feels that making them optional is maybe the best choice but rationally I'm having a hard time giving a clear example of why this is a must. 

There's a transformer in there called `AddRandom` which is supposed to add random vectors to the EmbeddingSet. Before, I would have argued that this is a use-case for which we must have the vectors added but now I wonder if there's maybe an easier way. You can also just create a random EmbeddingSet yourself and just merge it with the original one.

I suppose it feels "safe" to make them optional, mainly because there's a feeling of "I might be forgetting about something". Then again, I'm having a hard time coming up with a proper-usecase. 

Implementation wise, it'd be amazing to just be able to drop them all-together. This should vastly simplify the transformers because we can essentially wrap around the scikit-learn API.

mkaze:
[...]

PR #145, opened 2020-08-14
[WIP] Add support for integer axis as well as setting axis label and title in plot method

mkaze:
This PR addresses both issues #125 and #127 for matplotlib plots (i.e. `plot` method of `Embedding` and `EmbeddingSet`). The tests will be added.
Although, I am not satisfied with the way that `handle_2d_plot` function works; it only accepts an `Embedding` instance and therefore in `plot` method of `EmbeddingSet` things like determining axis is unnecessarily done over and over. I think there is room for improvement there. Further, this PR may address the issue #140.

**Backwards incompatible changes**:
- If `foo` is a 2D embedding, previously running `foo.plot(x_axis='xx', y_axis='yy')` was possible. Now, in `Embedding.plot` method, `x_axis` and `y_axis` can only take values of type integer or `Embedding`. In this case, to recreate the same plot as before, `x_label` and `y_label` arguments should be used instead.

**Bugs this PR resolve:**
- Previously, if `emb` was an embedding set (with dim > 2) and `'foo'` and `'bar'` were members of it, running `emb.plot(x_axis='foo', y_axis='bar')` would raise an error. This is resolved now.
- Previously if `foo`, `bar` and `buz` were all 2D embeddings, running `foo.plot(x_axis=bar, y_axis=buz)` would ignore `bar` and `buz` and plot the `foo` embedding itself (i.e. without projecting it onto `bar` and `buz`).

### DIFF
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -178,8 +178,11 @@ class Embedding:
     def plot(
         self,
         kind: str = "scatter",
-        x_axis: Union[str, "Embedding"] = None,
-        y_axis: Union[str, "Embedding"] = None,
+        x_axis: Union[int, "Embedding"] = None,
+        y_axis: Union[int, "Embedding"] = None,
+        x_label: Optional[str] = None,
+        y_label: Optional[str] = None,
+        title: Optional[str] = None,
         color: str = None,
         show_ops: bool = False,
         annot: bool = False,
@@ -190,8 +193,13 @@ class Embedding:
 
         Arguments:
             kind: what kind of plot to make, can be `scatter`, `arrow` or `text`
-            x_axis: the x-axis to be used, must be given when dim > 2
-            y_axis: the y-axis to be used, must be given when dim > 2
+            x_axis: the x-axis to be used, must be given when dim > 2; if an integer, the corresponding
+                dimension of embedding is used.
+            y_axis: the y-axis to be used, must be given when dim > 2; if an integer, the corresponding
+                dimension of embedding is used.
+            x_label: an optional label used for x-axis; if not given, it is set based on `x_axis` value.
+            y_label: an optional label used for y-axis; if not given, it is set based on `y_axis` value.
+            title: an optional title for the plot.
             color: the color of the dots
             show_ops: setting to also show the applied operations, only works for `text`
             annot: should the points be annotated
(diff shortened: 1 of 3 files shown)


koaning:
These changes look solid to me, the only thing that I am wondering is how we might be able to test this. 

One benefit of `Altair` is that you can serialise the entire chat to json which means that you can relatively easily test aspects of it. Matplotlib is a different story, but it's still a very important backend to get right. Especially with the planned changes, it might be time to consider what might be a good approach.

@mkaze got experience with https://github.com/matplotlib/pytest-mpl? I don't, but it might help out.

mkaze:
> got experience with https://github.com/matplotlib/pytest-mpl?

Yeah, I have already seen that but never worked with it so far. Actually, with that, there are four phases: 1) you write a test function which generates a plot, 2) run the tests once to produce the baseline images, 3) you manually check the generated baseline image files to make sure they are fine and store them in a directory in repository, 4) run the tests again, this time a new series of images is generated and then compared to the baseline images to make sure they are the same.

I would say this is a user-aided visual approach for testing the plot. However, there is another way of testing a matplotlib plot which is data-/props-based approach (similar to what you mentioned about altair) where you can check the information contained in the plot. In other words, in this approach we assume that functions such as `plt.plot` or `plt.scatter` works correctly; however, we want to make sure that the data and config passed to them were as they should be. I am currently more interested in this approach.

koaning:
You certainly are on to something there. I've tried to do similar things in the `EmbeddingSet` with methods like `to_X` and `to_dataframe` and we might be able to do something similar here. 

I wonder if there's a middle road possible in this instance. We could come up with different arguments for a `.plot()` call that should result in the same image. That should be relatively easy to test and covers a lot of ground. More of a smoketest than a unittest though.

koaning:
@mkaze just to check, is this PR still a work in progress?

mkaze:
@koaning Sorry for the delay! This week was a bit chaotic for me!
It's funny that I was working on it now and just made some commits right after your recent comment. :smile:

koaning:
No worries. It's all volunteer work. Good to know you're still working on it tho. 😄

mkaze:
@koaning The tests were added. Although, there is room for refactoring in the tests but I did not do it in this PR since it's a bit complicated. I think it's ready to be reviewed!

tests/test_plot.py
```diff
@@ -0,0 +1,415 @@
...
+
+def test_embedding_plot_raises_error_when_no_axis(embset):
+    emb = embset["red"]
+    with pytest.raises(ValueError, match="The `x_axis` value cannot be None"):
+        emb.plot()
+    with pytest.raises(ValueError, match="The `y_axis` value cannot be None"):
+        emb.plot(x_axis=0)
+
+
+## ===> EmbeddingSet.plot tests <============================================ ##
```
koaning, line 170: Something is telling me it might be best to split up the tests into two separate files. The test file is pretty big as is.

whatlies/embedding.py
```diff
@@ -212,29 +220,32 @@ def plot(
...
+            xlabel=x_label,
+            ylabel=y_label,
+            title=title,
             show_operations=show_ops,
             annot=annot,
             axis_option=axis_option,
         )
         return self
+
+    def _get_plot_axis_value_and_label(self, axis, dir):
```
koaning, line 241: You can still add a docstring to this method. Because there is an underscore at the start of the name, `mkdocs` will automatically understand that it should not appear in the API docs.

tests/test_plot.py
mkaze: Honestly, I considered that but ended up not doing it hesitantly :smile: I think I didn't do it because they were all about the `plot` method and also the `validation_plot_general_properties` was commonly used. Do you think it's better to split it, then?

koaning: In the other PR I've made a folder for testing visualisations. Probably best to separate the `matplotlib` based tests and the `altair` based tests. But given that we have a folder for `matplotlib` it feels preferable to add a small `common.py` where we might store useful functions that help us with tests. Scrolling all day can be a downer and if something goes wrong it can be productive if the tests files are all somewhat specific.

whatlies/embedding.py
koaning: I don't think the docstring needs to be as explicit as a one for a user-facing method. But it can be useful for internal use if there's a mention why the function was created.

koaning:
I need to go into a meeting now so I'll try to do some more reviewing later. I do want to mention: @mkaze +1 on the method of testing. It's the first time I've seen this but I think it's a perfect balance between "covering lots of ground" and "being able to test specific attributes". Was not aware that this was possible.

PR #147, opened 2020-08-15
Update Github Actions workflow files

mkaze:
This PR updates `croncheck` workflow as follow:

- Adds Python 3.6
- Removes OS specific installation
- Adds `en_core_web_sm` installation ([last check was failed due to this](https://github.com/RasaHQ/whatlies/runs/987110923))
- Removes Pytorch installation (also for `pythonpackage` workflow)
- Removes redundant `flake8` installation (also for `pythonpackage` workflow)

And also adds `torch` as a test dependency in `setup.py`.

### DIFF
--- a/.github/workflows/croncheck.yml
+++ b/.github/workflows/croncheck.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-10.15, ubuntu-latest]
-        python-version: [3.7]
+        python-version: [3.6, 3.7]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/croncheck.yml
```diff
@@ -10,30 +10,22 @@ jobs:
...
       uses: actions/setup-python@v1
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies Ubuntu
-      if: ${{matrix.os == 'ubuntu-latest'}}
+    - name: Install General Dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
         pip install -e ".[dev]"
         pip install flake8 pytest black
```
koaning, line 25: I think we're no longre installing `torch` here.

koaning:
Yeah this simplifies things again. LGTM!

PR #149, opened 2020-08-17
Version Bump

koaning:


koaning:
@mkaze I've just released this. We now have a release guide on the docs page too.

PR #157, opened 2020-08-24
Fix #154 - Add base class for transformers

### LINKED ISSUE #154: Parent class for transformers
@koaning Don't you think it's the time for a bit of refactoring and adding a parent `Transformer` class for all the transformers to subclass from it? For example, all of them have the same `__call__` method implementation. And also I think this makes the code more structured as it's getting bigger.

Willing to work on it if you are agreed.

mkaze:
This PR fixes #154 by adding an abstract base class for transformers and also performing some refactoring (i.e. removing `__call__` method implementation from subclasses).

### DIFF
--- a/whatlies/transformers/addrandom.py
+++ b/whatlies/transformers/addrandom.py
@@ -1,10 +1,11 @@
 import numpy as np
 
+from whatlies.transformers import Transformer
 from whatlies import Embedding, EmbeddingSet
 from whatlies.transformers.common import embset_to_X
 
 
-class AddRandom:
+class AddRandom(Transformer):
     """
     This transformer adds random embeddings to the embeddingset.
 
@@ -34,11 +35,6 @@ class AddRandom:
         self.sigma = sigma
         self.seed = seed
         self.is_fitted = False
-
-    def __call__(self, embset):
-        if not self.is_fitted:
-            self.fit(embset)
-        return self.transform(embset)
 
     def fit(self, embset):
         embset_to_X(embset=embset)
--- a/whatlies/transformers/addrandom.py
+++ b/whatlies/transformers/addrandom.py
@@ -31,10 +31,10 @@ class AddRandom(Transformer):
     """
 
     def __init__(self, n=1, sigma=0.1, seed=42):
+        super().__init__()
         self.n = n
         self.sigma = sigma
         self.seed = seed
-        self.is_fitted = False
 
     def fit(self, embset):
         embset_to_X(embset=embset)
--- a/whatlies/transformers/ivis.py
+++ b/whatlies/transformers/ivis.py
@@ -1,3 +1,4 @@
+from whatlies.transformers import Transformer
 from whatlies import EmbeddingSet
 from whatlies.transformers.common import embset_to_X, new_embedding_dict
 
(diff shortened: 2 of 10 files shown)


whatlies/transformers/pca.py
```diff
@@ -39,11 +40,6 @@ def __init__(self, n_components=2, **kwargs):
         self.kwargs = kwargs
```
koaning, line 40: Can't we remove the `self.is_fitted` here now?

whatlies/transformers/addrandom.py
```diff
@@ -35,11 +36,6 @@ def __init__(self, n=1, sigma=0.1, seed=42):
         self.seed = seed
         self.is_fitted = False
```
koaning, line 38: I think we can remove this `self.is_fitted` now.

whatlies/transformers/ivis.py
```diff
@@ -48,11 +49,6 @@ def __init__(self, n_components=2, **kwargs):
         self.kwargs["verbose"] = 0
         self.tfm = IVIS(embedding_dims=self.n_components, **self.kwargs)
```
koaning, line 50: Can we remove the `self.is_fitted` now in `__init__`?

whatlies/transformers/noise.py
```diff
@@ -38,11 +39,6 @@ def __init__(self, sigma=0.1, seed=42):
             lambda X: X + np.random.normal(0, sigma, X.shape)
```
koaning, line 39: Can we remove the `self.is_fitted` now in `__init__`?

whatlies/transformers/opentsne.py
```diff
@@ -51,12 +52,6 @@ def __init__(self, n_components=2, **kwargs):
         self.kwargs = kwargs
```
koaning, line 52: Can we remove the `self.is_fitted` now in `__init__`?

whatlies/transformers/tsne.py
```diff
@@ -43,11 +44,6 @@ def __init__(self, n_components=2, **kwargs):
         self.kwargs = kwargs
```
koaning, line 44: Can we remove the `self.is_fitted` now in `__init__`?

whatlies/transformers/umap.py
```diff
@@ -41,11 +42,6 @@ def __init__(self, n_components=2, **kwargs):
         self.kwargs = kwargs
```
koaning, line 42: Can we remove the `self.is_fitted` now in `__init__`?

koaning (changes requested):
There might be some more code that we can remove now. Other than that -> this is sweet. 

In the future we can probably even extend this by making a parent class for all the scikit-learn based transformers but this is a great first cleanup!

mkaze:
Actually, to remove the `self.is_fitted = False` in `__init__` method of subclasses they should call the base class constructure instead. I didn't do this and instead just added `self.is_fitted = False` to `__init__` method to make the (future!) developers aware that there is this property which should be maintained! I didn't like it even then, though! :smile: 

In other words, to initialize `self.is_fitted` either:

- `self.is_fitted = False` should be present in `__init__` method of all subclasses, or
- `super().__init__()` should be present in `__init__` method of all subclasses.

Of course, the second option is logical to me. @koaning Now, what do you think?

koaning:
@mkaze If we're going to be refactoring I'd say the `super().__init__()` is going to be the way forward.

koaning (approved):

koaning:
LGTM

PR #164, opened 2020-08-24
Fix #161,#162 - Add to_names_X and from_names_X methods to EmbeddingSet

### LINKED ISSUE #161: Another way of constructing an EmbeddingSet
Shouldn't we support initializing an `EmbeddingSet` using a list of names and an array/list/iterable of vectors:

```python
vecs = np.random.rand(3,4)
names = ['foo', 'bar', 'buz']

embset = EmbeddingSet(names, vecs)
# or
embset = EmbeddingSet.from_names_vecs(names, vecs)
```
Currently the most concise way of writing this is like: `EmbeddingSet(*[Embedding(n,v) for zip(names, vecs)])`. It sounds a common use case to me (for example in some of the transformers when used in `transform` method), but it shouldn't be that verbose to achieve! Preferably, it could also handle duplicate names by using the vector associated with the last occurrence of a name (which is automatically handled if we use something like `self.embeddings = {n:v for n, v in zip(names, vecs)}` in its implementation).

@koaning What do you think?

### LINKED ISSUE #162: A method to get names and vectors of an EmbeddingSet
Suggested:
```python
names, X = embset.to_names_vecs()
```
The most concise way to achieve it with current API:
```python
names = list(embset.embeddings.keys())
X = embset.to_X()
```
Besides its potential use case, this could replace the utility function `embset_to_X` which is used in transformers. 
Further, I have a feeling, which I don't know is good or not, that the need for the end-user to access `embeddings` property of `EmbeddingSet` should be minimal since it's mostly an implementation logic.

mkaze:
This PR fixes #161 and fixes #162.

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -298,7 +298,6 @@ class EmbeddingSet:
         Usage:
 
         ```python
-        from whatlies.embedding import Embedding
         from whatlies.embeddingset import EmbeddingSet
 
         names = ["foo", "bar", "buz"]
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -310,6 +310,10 @@ class EmbeddingSet:
         emb = EmbeddingSet.from_names_X(names, vecs)
         """
         X = np.array(X)
+        if len(X) != len(names):
+            raise ValueError(
+                f"The number of given names ({len(names)}) and vectors ({len(X)}) should be the same."
+            )
         return cls({n: Embedding(n, v) for n, v in zip(names, X)})
 
     def transform(self, transformer):
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -267,6 +267,52 @@ class EmbeddingSet:
         y = np.array([getattr(e, y_label) for e in self.embeddings.values()])
         return X, y
 
+    def to_names_X(self):
+        """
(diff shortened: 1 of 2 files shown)


whatlies/embeddingset.py
```diff
@@ -267,6 +267,51 @@ def to_X_y(self, y_label):
...
+        names = ["foo", "bar", "buz"]
+        vecs = [
+            [0.1, 0.3],
+            [0.7, 0.2],
+            [0.1, 0.9],
+        ]
+
+        emb = EmbeddingSet.from_names_X(names, vecs)
+        """
+        X = np.array(X)
```
koaning, line 312: I think we should test if `X.shape[0]` == `len(names)` here. Otherwise `zip` might cause unexpected behaviour.

koaning (changes requested):
I think we might need to check a shape, `zip` can cause unexpected behaviour. 

```
> list(zip('abc', [1,2]))
[('a', 1), ('b', 2)]
```

mkaze:
@koaning `zip` would consume the passed iterables until either of them have no elements and hence it would stop there. But that's a good think to notify the user by raising an error. I'll work on it.

sara-tagger:
Thanks for submitting a pull request 🚀 @TyDunn will take a look at it as soon as possible ✨

koaning (approved):

koaning:
LGTM!

PR #165, opened 2020-08-24
Fix #153 - Return self in fit method of transformers

### LINKED ISSUE #153: Return value of fit method in transformers
A suggestion: isn't it a good idea that the `fit` method in all the transformers return the object instance, i.e. `self`, so that we could chain `fit` and `transform` methods, like in:

```python
pca = Pca(...)
pca.fit(embset_1).transform(embset_2)
```
@koaning I can push a PR for that if you are agreed with this.

mkaze:
This PR fixes #153.


sara-tagger:
Thanks for submitting a pull request 🚀 @amn41 will take a look at it as soon as possible ✨

koaning (approved):

PR #166, opened 2020-08-25
Add missing spacy model to croncheck

mkaze:
The scheduled tests [are failing](https://github.com/RasaHQ/whatlies/actions?query=workflow%3A%22Cron+Test+Dependencies%22) due to this.


koaning:
That was also on my radar and you beat me to it. Thanks!

PR #168, opened 2020-08-25
Fix #150 - Add Normalizer transformer

### LINKED ISSUE #150: Add Transformer that Normalises Vectors
I'm currently adding a `norm` argument to the `plot_distance` method that normalises the vectors before being plotted. This is useful for plotting cosine distance. This made me consider that perhaps users might want to be able to do this anyway on the collection of embeddings. A transformer for this could help.

mkaze:
This PR fixes #150 by implementing `Normalizer` transformer for normalizing embeddings in an `EmbeddingSet` instance.

### DIFF
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -100,12 +100,12 @@ def test_normalizer_transformer():
     X = emb.to_X()
     normalized_emb = emb.transform(Normalizer())
     assert set(emb.embeddings.keys()) == set(normalized_emb.embeddings.keys())
-    assert np.array_equal(normalized_emb.to_X(), normalize(X))
-
-    normalized_emb = emb.transform(Normalizer(norm="l1"))
-    assert set(emb.embeddings.keys()) == set(normalized_emb.embeddings.keys())
     assert np.array_equal(normalized_emb.to_X(), normalize(X, norm="l1"))
+
+    normalized_emb = emb.transform(Normalizer(norm="l2"))
+    assert set(emb.embeddings.keys()) == set(normalized_emb.embeddings.keys())
+    assert np.array_equal(normalized_emb.to_X(), normalize(X))
 
     normalized_emb = emb.transform(Normalizer(feature=True))
     assert set(emb.embeddings.keys()) == set(normalized_emb.embeddings.keys())
-    assert np.array_equal(normalized_emb.to_X(), normalize(X, axis=0))
+    assert np.array_equal(normalized_emb.to_X(), normalize(X, norm="l1", axis=0))
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -1,6 +1,10 @@
 import pytest
+
+import numpy as np
+from sklearn.preprocessing import normalize
 from spacy.vocab import Vocab
 from spacy.language import Language
+
 from whatlies.language import SpacyLanguage
 from whatlies.transformers import (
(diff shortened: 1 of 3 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @koaning will take a look at it as soon as possible ✨

mkaze:
@sara-tagger Finally! You got it right this time :smile:

koaning:
Haha. Yeah we're trying to get that tagger to become slightly more intelligent. Internally though people seem to have learned that when they see "whatlies" that they just ping me on slack. :)

whatlies/transformers/normalizer.py
```diff
@@ -0,0 +1,50 @@
...
+
+    lang = SpacyLanguage("en_core_web_md")
+    emb = lang[words]
+
+    emb.transform(Normalizer(norm='l1'))
+    emb.transform(Normalizer(feature=True))
+    ```
+    """
+
+    def __init__(self, norm: str = "l2", feature: bool = False) -> None:
```
koaning, line 36: I think the most common way (considering the visualisation use-case) of normalising here is to "make everything sum to 1". With that in mind, might it make sense to us `l1` as a norm?

tests/test_transformers.py
```diff
@@ -85,3 +94,18 @@ def test_transformers_are_subclassed_properly(transformer):
 def test_transformer_base_class():
     with pytest.raises(TypeError, match="Can't instantiate abstract class Transformer"):
         Transformer()
+
+
```
koaning, line 98: Maybe add a test that confirms that when you just run the base normalise transformer that the sum of each vectors values is equal to one? Since that might be the main use-case?

whatlies/transformers/normalizer.py
mkaze: I am not sure, but if you say so I'll make the change.

tests/test_transformers.py
mkaze: Since we are comparing the results with `sklearn.preprocessing.normalize` in the tests and assuming that function is working correctly, I don't think there is a need to check the sums as well.

whatlies/transformers/normalizer.py
koaning: The reasoning is that cosine distances fit neatly in [0, 1] if the vectors sum up to one. That's how I realised that it deserves to be a feature.

tests/test_transformers.py
koaning: Fair enough.

whatlies/transformers/normalizer.py
mkaze: I made the change.
> The reasoning is that cosine distances fit neatly in [0, 1] if the vectors sum up to one.

Really?! Do you mean, in that case, the range of the range of cosine distance would be from 0 to 1 instead of -1 to 1?

koaning: d0h. This was my bad. If two 2D vectors are positive they are [0, 1] but you're completely right. In general it should be [0,2]. 

I still prefer `l1` because it feels easier to reason about. It simply sums to one. I'll gladly be convinced if for word embeddings you feel `l2` is better.

mkaze: I see, that's right. Generally, I have mostly used `l2` norm; but I don't have a specific preference here.

PR #173, opened 2020-08-25
Fix #122 - Add ndim attribute for Embedding and EmbeddingSet

### LINKED ISSUE #122: ndim attribute
@koaning Do you think it's a good idea to add a `ndim` (or `size`) attribute for `Embedding` class which gives the dimension of embedding vector? And what about the `EmbeddingSet`? For `EmbeddingSet` this makes sense only if all the embedding vectors in it have the same length (which I think is an implicit assumption so far; although, one might argue that this is not a necessary condition to hold for all applications).

mkaze:
This PR fixes #122 by adding a `ndim` attribute for `Embedding` class to represent the dimension of embedding vector.

@koaning I did not add this for `EmbeddingSet` since I think you were not in favor of it as discussed in #122. Anyways, the way to implement it would be:

```python
@property
def ndim(self):
    return self.embeddings.values()[0].ndim
```

### DIFF
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -76,3 +76,8 @@ def test_emb_plot_no_err_3d():
 def test_emb_str_method(emb):
     for char in "xyz":
         assert str(emb[char]) == char
+
+
+def test_emb_ndim():
+    foo = Embedding("foo", [0, 1, 0.2])
+    assert foo.ndim == 3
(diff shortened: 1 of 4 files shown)


koaning:
Feel free to also add `ndim` to the `EmbeddingSet` as well. We've got a checking mechanism now that checks that all dims should be the same and there's also something to be said for consistency.

koaning (approved):

PR #174, opened 2020-08-26
Fix #172 - Use `transform` method instead of `fit_transform` in Ivis transformer

### LINKED ISSUE #172: Using `fit_transform` in Ivis and Tsne transformers implementation
Shouldn't [here][1] and [here][2] the `transform` be used instead of `fit_transform`?! Otherwise it would refit on the (possibly new) given embeddingset and therefore produce wrong results; am I right?

[1]: https://github.com/RasaHQ/whatlies/blob/6100d271162297fc6876dc5dc17a8b56ddb1f173/whatlies/transformers/tsne.py#L55
[2]: https://github.com/RasaHQ/whatlies/blob/6100d271162297fc6876dc5dc17a8b56ddb1f173/whatlies/transformers/ivis.py#L60

mkaze:
This PR fixes #172.

### DIFF
--- a/whatlies/transformers/ivis.py
+++ b/whatlies/transformers/ivis.py
@@ -57,7 +57,7 @@ class Ivis(Transformer):
 
     def transform(self, embset):
         names, X = embset.to_names_X()
-        new_vecs = self.tfm.fit_transform(X)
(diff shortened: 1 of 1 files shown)


sara-tagger:

PR #176, opened 2020-08-26
Add Examples to the Documentation. 

koaning:
This PR will add an examples section to the documentation. That means that we'll have a place to put content that is a bit more "about a problem" than "about how the api works". The hope is that this can be an engaging way to get people to learn about the tool as well.


koaning:
@mkaze I think I've got a nice first example listed. To properly view it I'd suggest running it locally because there's a lot of diagrams explaining what is happening.

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
+In this example we'd like to discuss the effectiveness of debiasing techniques in word embeddings.
+This example is heavily inspired by the [lipstick on a pig paper](https://arxiv.org/pdf/1903.03862.pdf)
```
mkaze, line 2: It's better to be written as:

inspired by the \["Lipstick on a Pig"](https://arxiv.org/pdf/1903.03862.pdf) paper

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+by Hila Gonen and Yoav Goldberg.
+
+## Meaning
+
+In word embedding space you might wonder if a direction in embedding space might represents meaning.
```
mkaze, line 7: represent -> no "s" at the end because of using "might".

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+- $v_{man}$ represents the word embedding for "man" .
+- $v_{woman}$ represents the word embedding for "woman" .
+- You could argue that the axis $v_{man} - v_{woman}$ might represent "gender".
+
+One side of the "gender" axis would represent the male gender while the other end represents the female gender. This spectrum might allow you to guess if words are more "male" or more "female". It can also be used as a measurement for bias in language.
+
+## Similarity
+
+There are some axes that we might come up with that should be similar to this gender axis.
```
mkaze, line 17: It's better to end it like:

...this gender axis, like:

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+- $v_{he} - v_{she}$
+- $v_{king} - v_{queen}$
+- $v_{brother} - v_{sister}$
+
+It would be highly unfortunate though if the following pairs of words would display similarity.
```
mkaze, line 23: colon at the end instead of dot.

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+- $v_{nurse} - v_{physician}$
+- $v_{nurse} - v_{surgeon}$
+- $v_{nurse} - v_{doctor}$
+
+Being a woman should not imply that you are a nurse just like that being a man should not imply that you are a surgeon. It'd be a shame if we were using embeddings where such stereotypical bias present. Unfortunately, it's likely in the embeddings. Historically woman have not gotten the same opportunities as men. This is bound to be reflected on websites like Wikipedia which are a common source of data to train word embeddings. So let's make a similarity chart to confirm if this is the case.
```
mkaze, line 29: 1. "Being a woman should not imply that you are a nurse just like that being a man should not imply that you are a surgeon." --> isn't the opposite way of saying this the better: "Being a nurse should not imply that you are a woman..." because that's what we call the bias in our thinking, not the reverse?!

2. ...such stereotypical bias **is** present.
3. Historically**,**
4. woman --> women (after historically)

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+<details>
+  <summary><b>Word Pairs.</b></summary>
```
mkaze, line 32: The dot is redundant.

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+def make_correlation_plot(pairs, language_model, metric="cosine"):
+    axes = [calc_axis(p, language_model) for p in pairs]
+    emb_pairs = EmbeddingSet(*flatten(axes))
+    emb_pairs.plot_distance(metric=metric)
+
+make_correlation_plot(pairs=all_pairs, language_model=lang_ft)
+```
+
+This code generates a similarity chart for fasttext embeddings, shown below.
```
mkaze, line 77: Isn't this a "distance" chart according to the code? Although, distance is opposite of similarity!

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+![](imgs/bias.png)
+
+Notice, that we indeed see correlation. The "gender" direction seems to correlate with the "doctor-nurse" direction as well. This does not bode well.
```
mkaze, line 81: Is the correlation really that strong? Because, to me, doesn't look like so! Maybe I am looking at the wrong place! Specifically, are you referring to "she-he" vs. "nurse-physician" (or vs. "nurse-surgeon")? Because except those two, I don't see a strong correlation in terms of gender bias?!!

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+queen = Embedding("queen", [0.7, 0.9])
+
+man.plot(kind="arrow", color="blue")
+woman.plot(kind="arrow", color="red")
+king.plot(kind="arrow", color="blue")
+queen.plot(kind="arrow", color="red")
+
+(queen - king).plot(kind="arrow", color="pink", show_ops=True)
+(man | (queen - king)).plot(kind="arrow", color="pink", show_ops=True)
+plt.axis('off');
```
mkaze, line 106: Shouldn't the `matplotlib` import be included for this line?

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+```
+</details>
+
+![](imgs/logo.png)
+
+In this example you can see what might happen if we project $v_{man}$ away from the $v_{queen} - v_{king}$ axis we get a new vector $v_{man} | (v_{queen} - v_{king})$.
```
mkaze, line 112: "you can see that if we project ...."

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+You can even see that if in this 2D example;
```
mkaze, line 114: "You can even see in this 2D example we have:"

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+    # Plot the result.
+    emb_pairs.plot_distance(metric=metric)
+
+make_debias_correlation_plot(pairs=all_pairs, language_model=lang_ft)
+```
+
+We'll now display the "before" as well as "after" chart.
+
+![](imgs/before-after.png)
```
mkaze, line 146: Something strange is happening here because the "before" plot colors is much brighter than its previous counter-part (i.e. `bias.png`), so much so that I can almost sense the bias in the "before" plot (unlike the `bias.png` plot)??!!!

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+
+       False       0.97      0.73      0.83       131
+        True       0.68      0.96      0.79        78
+
+    accuracy                           0.81       209
+   macro avg       0.82      0.84      0.81       209
+weighted avg       0.86      0.81      0.82       209
+```
+
+We're using the same model as before, but now we're giving it a the debiased
```
mkaze, line 274: "a" is redundant.

docs/examples/lipstick-pig.md
```diff
@@ -0,0 +1,315 @@
...
+weighted avg       0.83      0.83      0.83       209
+```
+
+If we train a model on the debiased embeddings and also apply it to another
+debiased set we're able to get 83% of the cases right. We were hoping more
+around 50% here.
+
+## Conclusion
+
+If seems that after using linear projections as a debiasing technique that
```
mkaze, line 309: The second "that" is redundant.

mkaze:
@koaning Awesome! This is very nice to have.

docs/examples/lipstick-pig.md
koaning: I've replaced it with: 

> It's not a high correlation, but we'd prefer if it simply were zero.

docs/examples/lipstick-pig.md
koaning: Whoa! Well spotted. The first chart was wrong!

koaning: Fixed!

docs/examples/lipstick-pig.md
mkaze: Fine! But I hope this one is not confusing either for the reader (i.e. zero for correlation, not zero distance).

PR #182, opened 2020-08-31
Add support for integer axis, axis labels and title for `plot_interactive`

mkaze:
This PR addresses issues #125 and #127 for Altair plots, i.e. `plot_interactive`. The changes are all backwards compatible (including the axes labels, text of title as well as the logic for removing axes vectors; although, for the last one it might have side effects, e.g. an embedding with name "Dimension 0" would be removed if axis is set to 0).

@koaning It seems the `plot_interactive` is not tested at all in the test suit. So I just complied and didn't add any tests and instead manually tested the changes to make sure they are working. Do you have any specific plan for adding the tests for this method?

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -892,8 +892,11 @@ class EmbeddingSet:
 
     def plot_interactive(
         self,
-        x_axis: Union[str, Embedding],
-        y_axis: Union[str, Embedding],
+        x_axis: Union[int, str, Embedding],
+        y_axis: Union[int, str, Embedding],
+        x_label: Optional[str] = None,
+        y_label: Optional[str] = None,
+        title: Optional[str] = None,
         annot: bool = True,
         show_axis_point: bool = False,
         color: Union[None, str] = None,
@@ -902,8 +905,13 @@ class EmbeddingSet:
         Makes highly interactive plot of the set of embeddings.
 
         Arguments:
-            x_axis: the x-axis to be used, must be given when dim > 2
-            y_axis: the y-axis to be used, must be given when dim > 2
+            x_axis: the x-axis to be used, must be given when dim > 2; if an integer, the corresponding
+                dimension of embedding is used.
+            y_axis: the y-axis to be used, must be given when dim > 2; if an integer, the corresponding
+                dimension of embedding is used.
+            x_label: an optional label used for x-axis; if not given, it is set based on `x_axis` value.
+            y_label: an optional label used for y-axis; if not given, it is set based on `y_axis` value.
+            title: an optional title for the plot; if not given, it is set based on `x_axis` and `y_axis` values.
             annot: drawn points should be annotated
             show_axis_point: ensure that the axis are drawn
             color: a property that will be used for plotting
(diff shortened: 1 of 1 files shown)


koaning:
@mkaze this is certainly on the mental todo list but it feels best to add this as an issue so that we might discuss what features we'd like to test. Altair offers a relatively convenient API to test the properties of a chart.

mkaze:
@koaning Do you possibly have any idea why the tests failed? It seems it's an style check issue, however because of using `--exit-zero` flag all of them should be treated as just warning?! And the `black` does not raise any error locally! So I couldn't figure out what the problem is?!

koaning:
My guesstimate is that we're not specifying the target version. Looking at the [docs](https://github.com/psf/black) it seems that you can set what version of python you'd like to refer to as a benchmark. Quote; 

```
  -t, --target-version [py27|py33|py34|py35|py36|py37|py38]
                                  Python versions that should be supported by
                                  Black's output. [default: per-file auto-
                                  detection]
```

This might explain part of the behaviour. If you're running python 3.7 locally then the black run for python 3.6 might have a different opinion. I doubt this explains everything, but it's worth to add.

koaning:
It also seems like `black` recently had an [update](https://pypi.org/project/black/#history). It might be that some breaking changes got introduced. We currently run this line in the CI: 

```
pip install flake8 purest black
```

Which means that we're always using the latest version. We might want to consider not doing that anymore.

koaning:
Considering that a docs update is also failing https://github.com/RasaHQ/whatlies/pull/176/checks?check_run_id=1032428463 I'm inclined to think it is caused by the black update.

mkaze:
@koaning That's right! I can confirm that the `black` maintainers has released a new version with more opinionated formatting and therefore this breaks the style checks. However, personally I am not in favor of pinning `black` version because it's not a dependency per say, and also we want our code to be compatible and consistent with the latest black updates (probably that's why we are using it in the first place!).

mkaze:
@koaning I suggest to create a new PR addressing the latest changes suggested by `black` (and/or `flake8`), and then after merging that PR we can rebase our current PRs based on it. I probably can do that if you are busy now :)

koaning:
I was trying to do that right now actually, but there's something strange happening. 

It seems that the version of black that you have when you run `black .` is different than the one that the `pre-commit` hook fetches. It's a bummer because the changes that the black cli wants to introduce are now made undone immediately by the pre-commit hook. 

Have you had any luck?

mkaze:
@koaning Run `pre-commit autoupdate` to update the hooks to the latest version, then perform the commit.

koaning (approved):

whatlies/embeddingset.py
```diff
@@ -892,8 +892,11 @@ def plot_movement(
 
     def plot_interactive(
         self,
-        x_axis: Union[str, Embedding],
-        y_axis: Union[str, Embedding],
+        x_axis: Union[int, str, Embedding],
+        y_axis: Union[int, str, Embedding],
```
koaning, line 896: Don't we want to add a default value for these? Unless specified I can imagine a use-case to have `x_axis`=0 and `y_axis=1`. That way you don't have to fiddle around when you're using `Umap` or `Pca`.

mkaze: @koaning Actually, I had that in mind and wanted to create an issue for it since this is also a valid point for `plot` method as well (and even `Embedding.plot`). If you are OK with it, I'll address this in a separate PR with `x_axis=0` and `y_axis=1` as the default for all plot methods. Is that fine with you?

koaning: Ah yeah, that's a better idea. Makes the testing a lot clearer too.

koaning:
LGTM!

PR #186, opened 2020-09-01
Update Black

koaning:
This PR hopefully fixes the breaking style checking in the CI. @mkaze the `black autoupdate` command works like a charm!


mkaze:
@koaning Nice!
I think all of these changes were caused by the trailing commas that were already present in function signatures, function calls or lists, which in the recent version of `black` is assumed and enforced to be in multiple lines if a trailing comma is present; otherwise, the line-length limit has not been violated at all.

koaning:
Most likely. Either way -> I'm merging this.

PR #188, opened 2020-09-01
Change default axis value of plot methods

mkaze:
This PR changes the default value of `x_axis` and `y_axis` arguments in `plot` and `plot_interactive` methods.

**Backwards incompatible changes:**

- With 2D embeddings, when using `Embedding.plot` or `EmbeddingSet.plot` without setting `x_axis` and `y_axis`, the default axis labels would be "Dimension 0" and "Dimension 1" (instead of "x" and "y").
- With 2D embeddings, when using `Embedding.plot` or `EmbeddingSet.plot`, the `x_axis` and `y_axis` arguments no longer can take `None` value (the previous default value).

### DIFF
--- a/tests/test_plotting/test_matplotlib/test_embedding_plot.py
+++ b/tests/test_plotting/test_matplotlib/test_embedding_plot.py
@@ -156,9 +156,9 @@ def test_embedding_plot_arrow_emb_axis(embset):
     mpl.pyplot.close(fig)
 
 
-def test_embedding_plot_raises_error_when_no_axis(embset):
+def test_embedding_plot_raises_error_when_incorret_axis_type(embset):
     emb = embset["red"]
-    with pytest.raises(ValueError, match="The `x_axis` value cannot be None"):
-        emb.plot()
-    with pytest.raises(ValueError, match="The `y_axis` value cannot be None"):
-        emb.plot(x_axis=0)
+    with pytest.raises(ValueError, match="The `x_axis` value should be"):
+        emb.plot(x_axis=1.0)
+    with pytest.raises(ValueError, match="The `y_axis` value should be"):
+        emb.plot(y_axis="blue")
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -189,8 +189,8 @@ class Embedding:
     def plot(
         self,
         kind: str = "arrow",
-        x_axis: Union[int, "Embedding"] = None,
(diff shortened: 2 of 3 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @b-quachtran will take a look at it as soon as possible ✨

koaning:
Technically there's also `plot_interactive_matrix`. I think it's fine to skip for this PR, but I was wondering if you had an idea how we might want to handle it there.

koaning (approved):
There's a small discussion to have on the `plot_interactive_matrix` item these changes look good to me.

koaning:
I've added another issue on the documentation of this feature https://github.com/RasaHQ/whatlies/issues/189. In practice this is going to simplify a lot of code but the docs might not have this change in them yet.

mkaze:
@koaning Ah, yeah I forgot that method. Well, one idea that comes to my mind is to use the same default arguments there, i.e. 0 and 1, like this:

```python
def plot_interactive_matrix(self, *axes, ...):
    if len(axes) == 0:
        axes = [0, 1]
    # the rest ...
```

koaning:
@mkaze What you propose does not feel like it's the intended way you're supposed to create charts using the interactive matrix plotter but ... it certainly feels consistent. There's something to be said to keep the plotting methods as similar as possible to make it easier for folks to go-ahead and play. 

Feel free to add your proposal. Consistency is also a nice feature for this library. 

Also, @mkaze thanks! This is a nice PR to get in there.

mkaze:
> @mkaze What you propose does not feel like it's the intended way you're supposed to create charts using the interactive matrix plotter but ... it certainly feels consistent.

Why?! Are you referring to using 0 and 1 as the default axes which may not make sense a lot for a matrix plot? Or something else which I am not aware of?

koaning:
@mkaze I should have been more clear there. My bad! 

I was referring to the fact that your proposal only plots two dimensions which is not really what the method was built for. I could argue that for two dimensions you should maybe just go for the `plot_interactive` instead of the `plot_interactive_matrix`. That said, the interactive matrix is different than the other plotting methods in the sense that it's much harder to come up with a sensible default that will serve most people. So with these things mind, consistency seems like the best design choice here. 

I really do appreciate the consistency, despite 2D not really being use-case for the `plot_interactive_matrix`. More-over; assuming 3D also does not feel right. 

The main thing I want to communicate is that your default here is very sensible.

PR #191, opened 2020-09-02
Set default value of `axes` and add support for integer and Embedding axes in `plot_interactive_matrix`

mkaze:
This PR sets the default value of `axes` in `plot_interactive_matrix` method to `[0, 1]` as [discussed here](https://github.com/RasaHQ/whatlies/pull/188#issuecomment-685446261). Also, it adds the support for integer and `Embedding` type axes to this method (e.g. we can use `embset.plot_interactive_matrix(0, 'foo', bar)`). It was manually tested and worked fine.

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -1001,7 +1001,7 @@ class EmbeddingSet:
 
     def plot_interactive_matrix(
         self,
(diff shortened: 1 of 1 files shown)


koaning:
LGTM!

PR #192, opened 2020-09-03
Fix #129 - Add support for alternative axis metrics in `plot` method

### LINKED ISSUE #129: Support for alternative axis projection metrics in plots
Consider that `emb` is an embedding set and we run `emb.plot(x_axis='man', y_axis='woman')`. This would compute the normalized scalar projection (i.e. operator `>`) of every embedding in `emb` onto embeddings for `'man'` and `'woman'` so that they could be plotted in a 2D coordinate system. 

Now the question is: why we should only stick to normalized scalar projection? For example, one might want to use another measure such as cosine similarity or euclidean distance to plot the embeddings. 

- First of all, does this make sense? (to me, yes!)
- Secondly, is there currently an easy and built-in way to achieve this? (AFAIK, no!)
- Thirdly, is adding an argument to `plot` and `plot_interactive` methods to support this a good way? e.g. `emb.plot(x_axis='man', y_axis='woman', axis_metric='cosine')`.

mkaze:
This PR partially addresses #129 by adding support for custom axis metrics in matplotlib plots (i.e. `plot` method).

---
**TODO**:

- [X] Implement support for custom axis metric in `Embedding.plot` method.
- [X] Add tests for custom axis metric of `Embedding.plot` method.
- [X] Implement support for custom axis metric in `EmbeddingSet.plot` method.
- [X] Add tests for custom axis metric of `EmbeddingSet.plot` method.

---

### DIFF
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -1,7 +1,8 @@
-from typing import Union, Optional
+from typing import Union, Optional, Sequence, Callable
 from copy import deepcopy
 
 import numpy as np
+import scipy.spatial.distance as scipy_distance
 from sklearn.metrics import pairwise_distances
 
 from whatlies.common import handle_2d_plot
@@ -191,6 +192,7 @@ class Embedding:
         kind: str = "arrow",
         x_axis: Union[int, "Embedding"] = 0,
         y_axis: Union[int, "Embedding"] = 1,
+        axis_metric: Optional[Union[str, Callable, Sequence]] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
         title: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


sara-tagger:
Thanks for submitting a pull request 🚀 @tttthomasssss will take a look at it as soon as possible ✨

tttthomasssss (approved):
LGTM - awesome, thanks!

whatlies/embedding.py
```diff
@@ -249,21 +266,52 @@ def plot(
...
+                return lambda x, y: 1.0 - scipy_distance.cosine(x, y)
+            elif axis_metric == "cosine_distance":
+                return scipy_distance.cosine
+            elif axis_metric == "euclidean":
+                return scipy_distance.euclidean
+            else:
+                raise ValueError(
+                    f"The given axis metric is not supported: {axis_metric}"
+                )
+        elif callable(axis_metric):
```
koaning, line 312: TIL about `callable`.

koaning:
Nice! This is a really cool feature! 

I'll merge this today and I might just do a release as well. We've added a lot of features.

Issue #159, opened 2020-08-24
Allow for embset_deb = embset | embset_gender_axes


koaning:
If I want to apply the classical debiasing technique I can now run:

```python
embset_deb = embset | (embset['man'] - embset['woman'])
```

This is neat, but it unfortunately only works on a single embedding. I cannot do this: 

```python
gender_axes = EmbeddingSet(
  embset['man'] - embset['woman'],
  embset['king'] - embset['queen'],
  embset['doctor'] - embset['nurse']
)

embset_deb = embset | gender_axes
```

There are use-cases for projecting away from a plane as opposed to a single axis. Right now the only method we can offer for this debasing use-case is; 

```python
gender_axis = EmbeddingSet(
  embset['man'] - embset['woman'],
  embset['king'] - embset['queen'],
  embset['doctor'] - embset['nurse']
).average()

embset_deb = embset | gender_axis
```

I think it would be nice to also allow the projection of planes. I don't think we can do this for the other operators though. 

Also, I am now of the opinion that this, as a debiasing technique, is faulty. I would still like to feature it in this library though because it is easy to demonstrate why it is so.

mkaze:
So basically by `a | b` where both `a` and `b` are `EmbeddingSet` instances you want each embedding in `a` to be orthogonalized to all embeddings in `b`, right? If so, I think it could be done with Gram Schmidt approach.

koaning:
@mkaze precisely! This is what Gram-Schmidt should do. I don't know if numpy/scipy implements this.

mkaze:
@koaning I am not completely sure, but maybe QR decomposition might help which have been implemented in numpy ([here](https://numpy.org/doc/stable/reference/generated/numpy.linalg.qr.html)).

koaning:
I've written my own implementation of (a variant of) Gram-Schmidt before for [another project](https://scikit-lego.readthedocs.io/en/latest/fairness.html#Information-Filter). I recall trying the QR route but having a lot of trouble getting it to align with the axes that I was interested in. I'll give it another look, but a custom function could be fine as well here.

mkaze:
@koaning All right! Though, I have read before that there might be some numerical issues which needs to be taken care of if you want to implement GS in the straightforward way

koaning:
Since the repository moved to my personal account I'm moving this project into maintenance mode. This issue, while interesting, likely won't get the attention it needs in the future.

PR #193, opened 2020-09-03
Gram Schmidt on EmbeddingSet #159

koaning:
I've created a first implementation to fix https://github.com/RasaHQ/whatlies/issues/159. I'm kind of proud of the notation but I should double check the numerical accuracy. 

Internally I am using this idea.

```python
import numpy as np
from functools import reduce  
from whatlies import EmbeddingSet 

away_words = ['prince', 'princess', 'man', 'woman']
away_embs = [emb[w] for w in away_words]
for i in range(len(away_words)):
    away_embs[i] = reduce(lambda a, b: b | a, away_embs[:i+1])
```

This is what `away_embs` now looks like: 

```
[Emb[prince],
 Emb[(princess | prince)],
 Emb[(man | ((princess | prince) | prince))],
 Emb[(woman | ((man | ((princess | prince) | prince)) | ((princess | prince) | prince)))]]
```

These are all orthogonal vectors now that I need to project away from. This is now easy; 

```python
new_set = EmbeddingSet(self.embeddings.copy())
for e in orth_away:
    new_set = new_set | e
```

I'm happy that I've been able to re-use the notation to this extend but, again, I should now add some tests and see if I can confirm that we're still numerically sound.


whatlies/embeddingset.py
```diff
@@ -95,6 +95,28 @@ def __contains__(self, item):
         """
         return item in self.embeddings.keys()
 
+    def __getitem__(self, thing):
```
mkaze, line 98: Why the change in position of this method?!

koaning: All the dunder methods are now together. I allmost found myself re-implementing that method because it wasn't near the others can I didn't find it. With a similar philosophy all the plotting methods are together at the bottom of the object.

mkaze: Ah, I see. All right!

koaning:
@mkaze are you against merging this as is? I can still add some extra tests for the current implementation but I'd prefer to delay the QR implementation until I've found a nice use-case that allows us to test the numerical aspects.

mkaze:
@koaning I was just looking at it. Are you sure the implementation logic is correct?! It doesn't look like so to me! Besides, the test you have added has two problems:
- It doesn't actually test the method for the added feature (i.e. the case where the second operand is an `EmbeddingSet`).
- It doesn't check and validate that the logic is actually correct, i.e. whether the resulting embeddigns are orthogonal to the given embeddings (the second operand).

whatlies/embeddingset.py
```diff
@@ -178,7 +200,24 @@ def __or__(self, other):
...
         """
-        new_embeddings = {k: emb | other for k, emb in self.embeddings.items()}
+        if isinstance(other, Embedding):
+            new_embeddings = {k: emb | other for k, emb in self.embeddings.items()}
+        elif isinstance(other, EmbeddingSet):
+            # Apply Gram-Schmidt to project away from a hyperplane instead of an axis
+            # First, create orthogonal vectors that span the space to project away from
+            orth_away = [other[w] for w in other]
+            for i in range(len(orth_away)):
+                orth_away[i] = reduce(lambda a, b: b | a, orth_away[: i + 1])
```
mkaze, line 210: The import line for `reduce` is missing.

whatlies/embeddingset.py
```diff
@@ -178,7 +200,24 @@ def __or__(self, other):
         (emb | buz).plot(kind="arrow")
         ```
         """
-        new_embeddings = {k: emb | other for k, emb in self.embeddings.items()}
+        if isinstance(other, Embedding):
+            new_embeddings = {k: emb | other for k, emb in self.embeddings.items()}
+        elif isinstance(other, EmbeddingSet):
+            # Apply Gram-Schmidt to project away from a hyperplane instead of an axis
+            # First, create orthogonal vectors that span the space to project away from
+            orth_away = [other[w] for w in other]
```
mkaze, line 208: You probably meant `[w for w in other]`?!

whatlies/embeddingset.py
koaning: Are you sure? The tests wouldn't pass if it were and locally at least it seems to be imported on line 2.

koaning:
I've now added a test that checks for orthogonality.

whatlies/embeddingset.py
mkaze: Sorry! That's right. I thought we haven't used it already in this file.

mkaze:
```python
from whatlies.language import SpacyLanguage

lang = SpacyLanguage('en_core_web_md')
embset1 = lang[['blue', 'red', 'green', 'dog']]
embset2 = lang[['color', 'cat', 'meet', 'air']]
x = embset1 | embset2

for e1 in x:
    for e2 in embset2:
        print(e1.orig, e2.name, e1.vector @ e2.vector)

# Output:
blue color -1.0015886
blue cat -1.1149356
blue meet -0.1489037
blue air -4.491012e-07
red color -0.87258
red cat -1.0220132
red meet -0.1383523
red air -3.3294782e-08
green color -1.8166131
green cat -1.427654
green meet -0.16886657
green air -1.5558908e-07
dog color -0.7157535
dog cat -0.29572958
dog meet -0.021123573
dog air 1.2514647e-08
```
Shouldn't all the values be close to zero, or have I made a mistake somewhere?

koaning:
@mkaze I'd argue you're correct here! Well done! 

I'll add this as a test example. Interesting to see that it is having no trouble with "air". 

I am reminded of one thing though. Are you 100% sure that these vectors are linearly independant of eachother? spaCy sometimes (in an effort to save on memory) allocates the same vector to multiple tokens.

mkaze:
> Interesting to see that it is having no trouble with the last projection "air".

That's basically because the `air` is the last embedding in the embeddingset, and therefore `e` in the final execution of this line `new_set = new_set | e` is actually `air`.

> Are you 100% sure that these vectors are linearly independant of eachother?

I can confirm that none of the embeddings in `embset1` is equal to those in `embset2`.

koaning:
I've found a fix using QR. 

image

Will go head and use that implementation now.

mkaze:
@koaning I was wondering, in this problem, if there is a guarantee that the resulting embeddings could produce the same vector space as the original ones.

koaning:
@mkaze what do you mean with "produce same vector space"? The embeddings still have coordinates that are defined in 300 dimensions but after a projection I'd imagine there's less information in the vectors because we effectively project away information.

koaning:
If there's a matrix with `mat = [v1 v2 v3 v4]` and the vectors inside are all linearly independant. Once we project away `v4` then we get `mat_new = [v1|v4, v2|v4, v3|v4, v4|v4]`. 

This `mat_new` would have three linearly independant vectors where the original had 4.

mkaze:
@koaning The input and output vectors of Gram-Schmidt process generate the same vector space (or have the same linear span). This is an important property, otherwise there might be many ways to generate an orthogonal set of vectors from an initial set of vectors (e.g. always outputting the basis vectors of `R^n`).

Now my recent comment was regarding and pondering the fact that whether this property also holds (or must hold) in this problem as well, since this is a bit different problem than the problem of vector orthogonalization.

koaning:
Given an orthogonal set `embset_away` then currently our `embset | embset_away` seems like a `1-1` mapping. What I do not know for 100% sure is if our projections are "minimum distance". 

I'm pretty sure that I'm currently solving this problem:

image

I don't know if this approach also solves this one: 

image

PR #198, opened 2020-09-05
Fix #129 - Add support for alternative axis metrics in Altair plot methods

### LINKED ISSUE #129: Support for alternative axis projection metrics in plots

mkaze:
This PR partially addresses #129 by adding support for custom axis metric in Altair plot methods. To make the implementation easier, we added a new optional argument for `EmbeddingSet.compare_against` method to take custom functions.

**Update**: The `EmbeddingSet.compare_against` were refactored to use the already present argument for handling custom functions as well.

---
**TODO**:

- [X] Add support for axis metric in `EmbeddingSet.plot_interactive` method.
- [X] Add support for axis metric in `EmbeddingSet.plot_interactive_matrix` method.
- [X] Add tests for `EmbeddingSet.compare_against` method.
- [X] Run manual local tests.

---

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -203,9 +203,12 @@ class EmbeddingSet:
         new_embeddings = {k: emb >> other for k, emb in self.embeddings.items()}
         return EmbeddingSet(new_embeddings, name=f"({self.name} >> {other.name})")
 
-    def compare_against(self, other, mapping="direct"):
+    def compare_against(self, other, mapping="direct", func=None):
         if mapping == "direct":
-            return [v > other for k, v in self.embeddings.items()]
+            if func is None:
+                return [v > other for v in self.embeddings.values()]
+            else:
+                return [func(v.vector, other.vector) for v in self.embeddings.values()]
 
     def pipe(self, func, *args, **kwargs):
         """
@@ -912,6 +915,7 @@ class EmbeddingSet:
         self,
         x_axis: Union[int, str, Embedding] = 0,
         y_axis: Union[int, str, Embedding] = 1,
+        axis_metric: Optional[Union[str, Callable, Sequence]] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
         title: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


sara-tagger:

whatlies/embeddingset.py
```diff
@@ -203,9 +203,12 @@ def __rshift__(self, other):
...
         return EmbeddingSet(new_embeddings, name=f"({self.name} >> {other.name})")
 
-    def compare_against(self, other, mapping="direct"):
+    def compare_against(self, other, mapping="direct", func=None):
         if mapping == "direct":
-            return [v > other for k, v in self.embeddings.items()]
+            if func is None:
+                return [v > other for v in self.embeddings.values()]
+            else:
+                return [func(v.vector, other.vector) for v in self.embeddings.values()]
```
koaning, line 211: Might it be nicer to do something like; 

```
if callable(mapping):
    return [mapping(v.vector, other.vector) for v in self.embeddings.values()]
```

I think this was my original idea with that function as well.

koaning: Nitpick; I suppose a unit test for this method can't hurt.

koaning (changes requested):
Minor comment on `compare_against` method.

whatlies/embeddingset.py
mkaze: Ah, right. I didn't think about that. Made some changes and tests to address this.

koaning:
LGTM!

PR #205, opened 2020-09-08
Fix #199 - Remove helper vectors from transformers

### LINKED ISSUE #199: Handle helper vectors added by some of the transformer classes
There was this small discussion [here](https://github.com/RasaHQ/whatlies/issues/125#issuecomment-685452160) about how to handle helper vectors, such as `pca_0`, `pca_1` etc., added by some of the transformer classes. Now we can discuss it here in a dedicated issue to find out how we want to go about handling them.

@koaning I think both of us could not found a use case for them, especially after the integer axis support has been added. However, I think you might fear that something comes up in the future that proves the otherwise. Now, what's your final thought on this? If you prefer to make them optional, what's your suggested way of implementing it? Adding a keyword argument to `__init__`? And should they be added by default or not?

mkaze:
This PR fixes #199 by removing the helper vectors (e.g. `pca_0`, `pca_1`) from transformers.

---
**TODOs**:

- [X] Remove helper vectors from transformers implementation (`Pca`, `Umap`, `Ivis`, `Tsne` and `OpenTsne`).
- [X] Remove usage of helper vectors from documentation (in "readme.md" file, instead update the axis labels to be compatible with the existing image file).
- [X] Update interactive charts json data to reflect the label changes, and also remove helper vectors datapoints.

---
**Backwards incompatible changes:**
- Helper vectors have been removed from transformers and therefore any access to them after transformations would raise an error. This also applies to their usage in plot methods, e.g. `embset.transform(Pca(2)).plot_interactive('pca_0', 'pca_1')` would raise an error.


tests/test_transformers.py
```diff
@@ -42,7 +42,7 @@
     Normalizer(),
     Normalizer(feature=True),
 ]
-extra_sizes = [2, 3, 2, 3, 0, 0, 4, 1, 0, 2, 3, 2, 2, 3, 0, 0]
+extra_sizes = [0, 0, 0, 0, 0, 0, 4, 1, 0, 0, 0, 0, 0, 0, 0, 0]
```
koaning, line 45: To be honest. I'm open to deprecating `AddRandom`. It feels like that operation should be done via "make a new embeddingset with random vectors and concatenate". Might be best for another PR but wanted to mention it.

sara-tagger:

tests/test_transformers.py
mkaze: All right! Though, since I can't think of a good use case for it, I won't comment further.

Issue #200, opened 2020-09-07
3D matplotlib charts


koaning:
I'm not 100% sure about this feature but I wanted a place to discuss it. 

I recently made 3D charts to explain the geometry behind the debiasing technique and why I think it doesn't work as one might expect.

![image](https://user-images.githubusercontent.com/1019791/92369663-94ab5200-f0f9-11ea-853e-216664eebcff.png)

![image](https://user-images.githubusercontent.com/1019791/92369678-97a64280-f0f9-11ea-9b2c-2b86e0f73860.png)

I'm drawing over the 3D matplotlib charts here, but it definately seems like I'm able to tell a better story because of these charts. Might be good to start thinking about how we might be able to support them here.

mkaze:
@koaning In my opinion, one of the cornerstones for improvement of this library is its visualization capabilities (alongside two other cornerstones: 1- additional methods or modules for better/easier exploration and understanding, 2- well-written application guides). So I strongly support this one in general.

koaning:
What might be a nice name for such a method? 

```
embset.plot_3d()
embset.plot_3d_mpl()
```

The reason for bringing it up: there may be other backends that offer this as well.

koaning:
@mkaze I could start working on this since I've got the code ready anyway. Two questions;

1. What might be the appropriate method name `.plot_3d()`? 
2. The same question, but now keeping the unified API in mind.

mkaze:
> 1. What might be the appropriate method name `.plot_3d()`?

Yes, I think `plot_3d` is a good name.

> 2. The same question, but now keeping the unified API in mind.

Well, I was thinking about implementing this in the `plot` method as well by adding an additional `z_axis` argument which could be optional (with a default value of `None`, unlike `x_axis` and `y_axis`, to plot 2D charts by default). This is [consistent with matplotlib API](https://matplotlib.org/mpl_toolkits/mplot3d/tutorial.html) as well; but I can't foresee all the obstacles in this path without implementing it myself (though, I know it's possible to do!), and you might have a better picture as you have already implemented it. And further, this would be also useful if we decide in the future to have an API like `embset.plot.scatter`, `embset.plot.arrow`, `embset.plot.distance`, etc.

koaning:
This is where the details might get tricky. If you supply the `.plot()` method with three axes ... should we then make a "3d"-plot or a "scatter-matrix"-plot like `.plot_interactive_matrix`? 

[...]

Issue #202, opened 2020-09-07
Visualization API and its connection to backends


mkaze:
A more or less recurring topic in the issues (and probably in our minds) is the discussion of visualization API and its connection to different visualization backends which are already supported or will be supported in the future. Hence, I thought it's worth having a dedicated issue for it in order to discuss it clearly and coherently in one place.

mkaze:
There is a relevant question that is lingering in my mind for a while: is there any way to have a unified visualization **public** API instead of having a separate method for each backend? In other words, instead of having `plot` and `plot_interactive` (or `plot_altair` vs. `plot_bokeh` vs. `plot_matplotlib` as suggested in the above comment), could we have a single `plot` method and handle backends differently under-the-hood? (of course, one may argue that there is another question which should be answered first: why does having a unified API matter? or how does it help?)

**Suggestion:** I think it could be implemented in one of the following ways or even a hybrid of them:
- Set backend globally. For example:
    ```python
    import whatlies

    # matplotlib will be used by default
    embset.plot(....)
    embset.plot_distance(...)    

    whatlies.set_vis_backend('altair')
    # all the following plots will be plotted by altair (if possible)
    embset.plot(...)
    embset.plot_distance(...) 
    ```

- Pass the desired visualization backend as an argument to the unified API methods. For example:
    ```python
    embset.plot(...)  # matlplotlib will be used by default
    embset.plot(..., backend="altair")    # altair will be used for this plot
    ```
As I mentioned above either one of these solutions, or preferably both of them could be implemented. I am not sure and convinced about the usefulness of these solutions, though. They may not save the user from writing, but at least it makes the API simple and unified. @koaning Since I read your comment [here](https://github.com/RasaHQ/whatlies/issues/200#issuecomment-688293664), I thought it would be useful to free my mind and write down this idea here :smile:

koaning:
I really like this proposal. It is breaking, but since we can keep `plot_interactive` around for a deprecation period I only see upsides. 

I would be interested in a hybrid solution of your proposal. I can imagine that certain charts work in inter-active mode while other charts are just better looking with matplotlib. For example; plot_distance is possible in altair, but perhaps not better looking. I can also imagine, with the advent of 3d plotting, that certain backends may not allow for everything. Matplotlib might allow for a 3d plot but Altair certainly does not.  Matplotlib allows you to draw arrows, I'm not 100% sure of Altair here.

mkaze:
[...]

PR #208, opened 2020-09-09
Adding 3D-Plot #200

koaning:
This adresses https://github.com/RasaHQ/whatlies/issues/200. 

```python
emb.transform(Pca(3)).plot_3d(annot=True)
```

image

```python
emb.transform(Pca(3)).plot_3d("king", "dude", "green", annot=True)
```

image

There's a few things about this PR worth discussing. 

1. The color property works differently in 3d charts. It seems as if the color is automatically altered depending on the depth of the point. Since this is a sensible way of dealing with it I'm wondering if this is something we want to influence directly. 
2. I'm currently adding an absolute value to offset the text from the 3d point. It kind of works but it depends on the zoom level. @mkaze might this be better as an argument? 
3. It feels like we might be able to generalise the plots. They always have a phase where a DataFrame is created for the plot and it seems like a lot of code is repeating itself. This is not something I want to adress in this PR but it is something I've observed. 
4. Testing 3D plots is different. When you call `ax.collections[0].get_offsets()` then you seem to get the 2d coordinates of the points that are plotted instead of the original 3d ones. I've added some basic tests for now, but I'll gladly hear it if we need more. I'm still investigating how the 3d properties work exactly in matplotlib.

### DIFF
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -301,7 +301,7 @@ class Embedding:
         elif isinstance(axis_metric, str):
             if axis_metric == "cosine":
                 raise ValueError(
-                    "Please be specific, do you want `cosine_distance` or `cosine_similarity`."
+                    "Please be specific, do you want `cosine_distance` or `cosine_similarity`?"
                 )
             elif axis_metric == "cosine_similarity":
                 return lambda x, y: 1.0 - scipy_distance.cosine(x, y)
--- a/whatlies/embedding.py
+++ b/whatlies/embedding.py
@@ -299,7 +299,11 @@ class Embedding:
         if axis_metric is None:
             return None
         elif isinstance(axis_metric, str):
-            if axis_metric == "cosine_similarity":
+            if axis_metric == "cosine":
+                raise ValueError(
+                    "Please be specific, do you want `cosine_distance` or `cosine_similarity`."
+                )
+            elif axis_metric == "cosine_similarity":
                 return lambda x, y: 1.0 - scipy_distance.cosine(x, y)
             elif axis_metric == "cosine_distance":
                 return scipy_distance.cosine
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -684,8 +684,12 @@ class EmbeddingSet:
         x_axis: Union[int, str, Embedding] = 0,
         y_axis: Union[int, str, Embedding] = 1,
         z_axis: Union[int, str, Embedding] = 2,
+        x_label: Optional[str] = None,
+        y_label: Optional[str] = None,
+        z_label: Optional[str] = None,
+        title: Optional[str] = None,
         color: str = None,
-        axes_metric: Optional[Union[str, Callable, Sequence]] = None,
+        axis_metric: Optional[Union[str, Callable, Sequence]] = None,
         annot: bool = True,
     ):
         """
@@ -698,12 +702,17 @@ class EmbeddingSet:
                 dimension of embedding is used.
             z_axis: the z-axis to be used, must be given when dim > 3; if an integer, the corresponding
                 dimension of embedding is used.
+            x_label: an optional label used for x-axis; if not given, it is set based on value of `x_axis`.
+            y_label: an optional label used for y-axis; if not given, it is set based on value of `y_axis`.
+            z_label: an optional label used for z-axis; if not given, it is set based on value of `z_axis`.
+            title: an optional title for the plot.
             color: the property to user for the color
-            axes_metric: the metric used to project each embedding on the axes; only used when the corresponding
+            axis_metric: the metric used to project each embedding on the axes; only used when the corresponding
                 axis is a string or an `Embedding` instance. It could be a string (`'cosine_similarity'`,
                 `'cosine_distance'` or `'euclidean'`), or a callable that takes two vectors as input and
-                returns a scalar value as output. To set different metrics for different axes, a list or a tuple of
-                the same length as `axes` could be given. By default (`None`), normalized scalar projection (i.e. `>` operator) is used.
+                returns a scalar value as output. To set different metrics of the three different axes,
+                a list or a tuple of the same length as `axes` could be given. By default (`None`),
+                normalized scalar projection (i.e. `>` operator) is used.
             annot: drawn points should be annotated
 
         **Usage**
@@ -730,14 +739,14 @@ class EmbeddingSet:
         if isinstance(z_axis, str):
             z_axis = self[z_axis]
 
-        if isinstance(axes_metric, (list, tuple)):
-            x_axis_metric = axes_metric[0]
-            y_axis_metric = axes_metric[1]
-            z_axis_metric = axes_metric[2]
+        if isinstance(axis_metric, (list, tuple)):
+            x_axis_metric = axis_metric[0]
+            y_axis_metric = axis_metric[1]
+            z_axis_metric = axis_metric[2]
         else:
-            x_axis_metric = axes_metric
-            y_axis_metric = axes_metric
-            z_axis_metric = axes_metric
+            x_axis_metric = axis_metric
+            y_axis_metric = axis_metric
+            z_axis_metric = axis_metric
 
         # Determine axes values and labels
         if isinstance(x_axis, int):
@@ -746,7 +755,7 @@ class EmbeddingSet:
         else:
             x_axis_metric = Embedding._get_plot_axis_metric_callable(x_axis_metric)
             x_val = self.compare_against(x_axis, mapping=x_axis_metric)
-            x_lab = x_axis.name
+            x_lab = x_label if x_label else x_axis.name
 
         if isinstance(y_axis, int):
             y_val = self.to_X()[:, y_axis]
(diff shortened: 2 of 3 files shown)


whatlies/embeddingset.py
```diff
@@ -679,6 +679,119 @@ def plot(
...
+            z_axis: the z-axis to be used, must be given when dim > 3; if an integer, the corresponding
+                dimension of embedding is used.
+            color: the property to user for the color
+            axes_metric: the metric used to project each embedding on the axes; only used when the corresponding
+                axis is a string or an `Embedding` instance. It could be a string (`'cosine_similarity'`,
+                `'cosine_distance'` or `'euclidean'`), or a callable that takes two vectors as input and
+                returns a scalar value as output. To set different metrics for different axes, a list or a tuple of
+                the same length as `axes` could be given. By default (`None`), normalized scalar projection (i.e. `>` operator) is used.
+            annot: drawn points should be annotated
+
```
mkaze, line 708: What about adding `x_label`, `y_label`, `z_label` and `title` arguments? For the sake of consistency!

whatlies/embedding.py
```diff
@@ -299,7 +299,11 @@ def _get_plot_axis_metric_callable(axis_metric) -> Optional[Callable]:
         if axis_metric is None:
             return None
         elif isinstance(axis_metric, str):
-            if axis_metric == "cosine_similarity":
+            if axis_metric == "cosine":
+                raise ValueError(
+                    "Please be specific, do you want `cosine_distance` or `cosine_similarity`."
```
mkaze, line 304: Since it's a question, a question mark should be put at the end instead.

whatlies/embeddingset.py
```diff
@@ -679,6 +679,119 @@ def plot(
...
+            y_axis: the y-axis to be used, must be given when dim > 3; if an integer, the corresponding
+                dimension of embedding is used.
+            z_axis: the z-axis to be used, must be given when dim > 3; if an integer, the corresponding
+                dimension of embedding is used.
+            color: the property to user for the color
+            axes_metric: the metric used to project each embedding on the axes; only used when the corresponding
+                axis is a string or an `Embedding` instance. It could be a string (`'cosine_similarity'`,
+                `'cosine_distance'` or `'euclidean'`), or a callable that takes two vectors as input and
+                returns a scalar value as output. To set different metrics for different axes, a list or a tuple of
+                the same length as `axes` could be given. By default (`None`), normalized scalar projection (i.e. `>` operator) is used.
```
mkaze, line 706: "To set different metrics for each of the three axes, a list or tuple of three elements could be given."

The current one is the one used for `plot_interactive_matrix`.

whatlies/embeddingset.py
```diff
@@ -679,6 +679,119 @@ def plot(
         )
         return self
 
+    def plot_3d(
+        self,
+        x_axis: Union[int, str, Embedding] = 0,
+        y_axis: Union[int, str, Embedding] = 1,
+        z_axis: Union[int, str, Embedding] = 2,
+        color: str = None,
+        axes_metric: Optional[Union[str, Callable, Sequence]] = None,
```
mkaze, line 688: To make it consistent with `plot` and `plot_interactive` let's use `axis_metric` instead of `axes_metric`.
I used `axes_metric` in `plot_interactive_matrix` because there was a `axes` argument. Maybe I shouldn't have done that!

mkaze:
> 1. The color property works differently in 3d charts. It seems as if the color is automatically altered depending on the depth of the point. Since this is a sensible way of dealing with it I'm wondering if this is something we want to influence directly.

Well, actually this could be controlled by `depthshade` argument of `scatter3D` which is `True` by default (see [documentation](https://matplotlib.org/3.1.1/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.html?highlight=scatter3d#mpl_toolkits.mplot3d.axes3d.Axes3D.scatter3D)); so this is something that could be controlled. Actually, in order to not provide a lot of custom arguments and at the same time give more flexibility to the user, I am thinking of adding a `**kwargs` argument for all plot methods where the user can provide additional arguments which could be potentially passed to the main underlying plot method (e.g. in this case it would be `scatter3D`) and we can clarify this in the docstrings. If you see the documentation, you will see that this is something that matplotlib does as well. However, I am not suggesting that this should be done in this PR.

> 2. I'm currently adding an absolute value to offset the text from the 3d point. It kind of works but it depends on the zoom level. @mkaze might this be better as an argument?

Well, I think `0.2` is a bit high, especially if the coordinate values is in the order of less than 1. `0.05` might be a better option. But, I am not sure if it's a good idea to control this using a separate argument.  

> 3. It feels like we might be able to generalise the plots. They always have a phase where a DataFrame is created for the plot and it seems like a lot of code is repeating itself. This is not something I want to adress in this PR but it is something I've observed.

That's right. There is a data preparation stage with some common steps in most of the plot methods. There is a potential to do some refactoring there (even for the methods that currently use `handle_2d_plot`). Though, this also could be addressed when refactoring the visualization API (#202) as well.

> 4. Testing 3D plots is different. When you call `ax.collections[0].get_offsets()` then you seem to get the 2d coordinates of the points that are plotted instead of the original 3d ones. I've added some basic tests for now, but I'll gladly hear it if we need more. I'm still investigating how the 3d properties work exactly in matplotlib.

First, I thought there must be some 3D to 2D projection is happening. So I looked in the methods list and found the `get_offset_transform`. However, although it seems to have some kind of projection matrix but none of its transform method worked as I expected. 
Next, I searched a bit and found [this answer](https://stackoverflow.com/a/51716781/2099607) on SO. This is basically what you want, however in the current implementation of `plot_3d` the return value of `scatter3D` is not returned (instead axis value is returned which seems to be a bit different). So if we are interested in automated testing of this plot, either the return value should be changed or we should continue searching for a workaround. My opinion is that to use the return value of `Scatter3D` as the return value of the method because then `.axes` property of that value would also gives us the `plt.axes(...)` return value.

whatlies/embeddingset.py
```diff
@@ -746,23 +755,23 @@ def plot_3d(
         else:
             x_axis_metric = Embedding._get_plot_axis_metric_callable(x_axis_metric)
             x_val = self.compare_against(x_axis, mapping=x_axis_metric)
-            x_lab = x_axis.name
+            x_lab = x_label if x_label else x_axis.name
```
mkaze, line 758: This line should be put after the `if/else`, otherwise it won't be effective when `if` condition is True. Further it should be like `x_lab = x_label if x_label is not None else ....`, otherwise when `x_label = ''` this would not be effective (someone may want to not use a label for an axis using this trick). The same goes for labels of other axes.

whatlies/embeddingset.py
```diff
@@ -698,12 +702,17 @@ def plot_3d(
...
+            title: an optional title for the plot.
             color: the property to user for the color
-            axes_metric: the metric used to project each embedding on the axes; only used when the corresponding
+            axis_metric: the metric used to project each embedding on the axes; only used when the corresponding
                 axis is a string or an `Embedding` instance. It could be a string (`'cosine_similarity'`,
                 `'cosine_distance'` or `'euclidean'`), or a callable that takes two vectors as input and
-                returns a scalar value as output. To set different metrics for different axes, a list or a tuple of
-                the same length as `axes` could be given. By default (`None`), normalized scalar projection (i.e. `>` operator) is used.
+                returns a scalar value as output. To set different metrics of the three different axes,
+                a list or a tuple of the same length as `axes` could be given. By default (`None`),
```
mkaze, line 714: As I mentioned in my previous comment, there is no "\`axes\`" for this method!

koaning:
@mkaze I've made some of the changes. 

- I've added `title` and `x/y/z_label` arguments. 
- I'm using the `ax.collections[0]._offsets3d` to test some of the drawn points. 
- Renamed some `axes`-arguments to `axis`-arguments. I think I've got all of them now. 
- I've added support for the `color` property which wasn't in there before. I'm not touching the depth, just allowing the user to pass a property like in the other methods for consistency. 

The title looks a bit ... strange in a 3d plot. But for consistency it's been added. 

```python
plt.figure(figsize=(7,7))
embset = emb.add_property("group", lambda d: random.random()).transform(Pca(3))
ax = embset.plot_3d(annot=True, title="foobar", color='group')
```

image

```python
plt.figure(figsize=(7,7))
embset = emb.add_property("group", lambda d: random.random() > 0.5).transform(Pca(3))
embset.plot_3d(annot=True, title="foobar", color='group')
```

image


My only caveat in using this method right now is the manual figsize that I am setting. With setting it, this is what the last chart looks like; 

image

The shape is a bit clumped together and you can also see that there is some part of the text missing on the right.

mkaze:
@koaning All right! Thanks for the changes.
Just one note about the `color` argument: how are we supposed to assign a single color to the points without using a custom property? For example, `plot_3d(...., color='red')`? I am asking this because the behavior of `color` argument in Altair plot methods is also the same.

koaning:
So currently if you pass `color='red'` to `.plot_interactive` then it kind of gets ignored. 

image

A similar thing happens in `plot_3d`. 

image

Effectively there's an internal column column with only `""` values that are made as a placeholder and it will be using this column to assign the color using what-ever the plotting backend is. I'm wondering what the better behavior is in this scenario. 

1. We might want to take the matplotlib standard and allow for a subset of strings that represent colors to change the colors of all the points. This is a standard in matplotlib, but other libraries deal with this differently.
2. We might throw an error instead, indicating that you can only pass a reference to a column name with the `color` argument. It won't be flexible, but might be easier to be consistent across different types of plots. 

We might also defer this decision to another PR since it this topic spans not just the `.plot3d()` method.

koaning:
Other than that, this good to merge @mkaze?

mkaze (approved):
It looks good to me. Thanks!

PR #212, opened 2020-09-10
Remove redundancies from altair plot methods

mkaze:
This PR removes some of the redundancies in altair plot methods when annotating data points with texts.


koaning (approved):

koaning:
LGTM.

Issue #155, opened 2020-08-23
Redo `add_property`


koaning:
Currently [the docs](https://rasahq.github.io/whatlies/tutorial/transformations/) describe how to add properties to an `EmbeddingSet` which can later be used for plotting. 

```python
from whatlies.language.language import SpacyLanguage
lang = SpacyLanguage('en_core_web_sm')

words1 = ["dog", "cat", "mouse", "deer", "elephant", "zebra", "fish",
          "rabbit", "rat", "tomato", "banana", "coffee", "tea", "apple", "union"]
words2 = ["run", "swim", "dance", "sit", "eat", "hear", "look", "run", "stand"]

umap = Umap(2)
emb1 = lang[words1].transform(umap).add_property('set', lambda d: 'set-one')
emb2 = lang[words2].transform(umap).add_property('set', lambda d: 'set-two')

both = emb1.merge(emb2)
```

I think we can do slightly better here. This would be much nicer: 

```python
umap = Umap(2)
emb1 = lang[words1].transform(umap).add_property(set=lambda d: 'set-one')
emb2 = lang[words2].transform(umap).add_property(set=lambda d: 'set-two')
```

The benefit of this is that you can add multiple properties in one call. The usecase here is that this allows you to run a clustering algorithm on the dataset and then attach the cluster-id as a property on each embedding. 

Since this is a breaking change I would propose that we might deprecate `add_property` in favor of  another name. But first I'd like to check what other people might think.  @mkaze ?

mkaze:
> The usecase here is that this allows you to run a clustering algorithm on the dataset and then attach the cluster-id as a property on each embedding.

Funny! I just saw this after mentioning the same thing in the other issue :smile: 

> Since this is a breaking change I would propose that we might deprecate add_property in favor of another name. But first I'd like to check what other people might think. @mkaze ?

I am in favor of the whole idea; however, I am not sure whether we should use a new method with a new name for that, or just make a backwards-incompatible change! Either has its own pros/cons.

koaning:
I might argue that `assign` can also be seen as a nicer "verb" here. Mainly because this is the method that pandas uses for a very similar operation.

PR #214, opened 2020-09-11
Add .assign verb #155

koaning:
This PR fixes https://github.com/RasaHQ/whatlies/issues/155. I've left the `add_property` intract since it's not hurting anyone but I have changed the internals of `embedding` slighty. I added a `copy()` command such that we have a convenient method to get a copy of the embedding that also has all of the properties attached. 

The api currently works swell but this PR is making me wonder if in the future, we may want to not have properties added to the `Embedding` interactively since it might be more used friendly to add all of these custom properties in a `data` field on the `Embedding`. Technically, currently a user can `embset.assign(vector=lambda d: 1)` and it'd be no bueno. 

@mkaze got thoughts on this? I'll gladly make an issue for this if you feel the same way.

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -446,6 +446,38 @@ class EmbeddingSet:
         """
         return EmbeddingSet({**self.embeddings, **other.embeddings})
 
+    def assign(self, **kwargs):
+        """
+        Adds properties to every embedding in the set based on the keyword arguments.
+
+        This is very useful for plotting because a property can be used to assign colors. This method is very
+        similar to `.add_property` but it might be more convenient when you want to assign multiple properties
+        in one single statement.
+
+        Arguments:
+            kwargs: (name, func)-pairs that describe the name of the property as well as a function on how to calculate it
+
+        Usage:
+
+        ```python
+        from whatlies.embeddingset import EmbeddingSet
+
+        foo = Embedding("foo", [0.1, 0.3, 0.10])
+        bar = Embedding("bar", [0.7, 0.2, 0.11])
+        emb = EmbeddingSet(foo, bar)
+        emb_with_property = emb.assign(dim0=lambda d: d.vector[0],
+                                       dim1=lambda d: d.vector[1],
+                                       dim2=lambda d: d.vector[2])
+        ```
+        """
+        new_set = {}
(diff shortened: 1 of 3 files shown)


mkaze:
> @mkaze got thoughts on this? I'll gladly make an issue for this if you feel the same way.

I also had the same idea, but forgot to mention it :smile: I even considered using `__getattr__` method so that we can still use the dot notation for accessing the attributes/properties which are instead stored in a specific data-structure (like the `data` you mentioned).

koaning:
@mkaze is this good to merge?

whatlies/embeddingset.py
```diff
@@ -446,6 +446,38 @@ def merge(self, other):
...
+        bar = Embedding("bar", [0.7, 0.2, 0.11])
+        emb = EmbeddingSet(foo, bar)
+        emb_with_property = emb.assign(dim0=lambda d: d.vector[0],
+                                       dim1=lambda d: d.vector[1],
+                                       dim2=lambda d: d.vector[2])
+        ```
+        """
+        new_set = {}
+        for k, e in self.embeddings.items():
+            new_emb = e.copy()
```
mkaze, line 475: I think this would be redundant since `add_property` returns a copy of the embedding per your recent change.

whatlies/embeddingset.py
```diff
@@ -446,6 +446,38 @@ def merge(self, other):
...
+    def assign(self, **kwargs):
+        """
+        Adds properties to every embedding in the set based on the keyword arguments.
+
+        This is very useful for plotting because a property can be used to assign colors. This method is very
+        similar to `.add_property` but it might be more convenient when you want to assign multiple properties
+        in one single statement.
+
+        Arguments:
+            kwargs: (name, func)-pairs that describe the name of the property as well as a function on how to calculate it
```
mkaze, line 458: I am not sure if it should be mentioned that the `func` is a function **that takes an `Embedding` instance as input**?!

mkaze (approved):
Besides those two minor comments, it looks good to me. Thanks!

whatlies/embeddingset.py
koaning: Explicit > Implicit.

Issue #183, opened 2020-08-31
Add tests for the interactive charts


koaning:
@mkaze recently introduced some tests for matplotlib based charts, it's time that we also do that for the ones based on Altair. This issue is a place where we can discuss what properties to test.

mkaze:
I'll take this one! :)

koaning:
@mkaze can I close this issue?

mkaze:
@koaning Actually, I haven't added the tests for `plot_movement` yet.

PR #215, opened 2020-09-11
Add tests for Altair plot methods, `plot_interactive` and `plot_interactive_matrix`

mkaze:
This PR partially addresses #183 by adding tests for `plot_interactive` and `plot_interactive_matrix`. The `plot_movement` method would be addressed in another PR because the method needs some changes.

---
**TODOs:**
- [X] Add tests for `plot_interactive` method.
- [X] Add tests for `plot_interactive_matrix` method.

---

### DIFF
--- a/tests/test_plotting/test_altair/common.py
+++ b/tests/test_plotting/test_altair/common.py
@@ -1,0 +1,10 @@
+def validate_plot_general_properties(chart_layer, props):
+    assert chart_layer["encoding"]["x"]["axis"]["title"] == props["x_label"]
+    assert chart_layer["encoding"]["y"]["axis"]["title"] == props["y_label"]
+    assert chart_layer["title"] == props["title"]
+    assert chart_layer["mark"]["type"] == props["type"]
+
+
+def validate_matrix_plot_general_properties(spec, props):
+    assert spec["width"] == props["width"]
+    assert spec["height"] == props["height"]
--- a/tests/test_plotting/test_altair/test_embeddingset_plot_interactive.py
+++ b/tests/test_plotting/test_altair/test_embeddingset_plot_interactive.py
@@ -1,0 +1,366 @@
+import json
+
+import pytest
+import numpy as np
+import pandas as pd
+import scipy.spatial.distance as scipy_distance
+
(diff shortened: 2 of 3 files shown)


koaning:
I'll need to properly sit down to review this one. Will probably get to that later this week.

mkaze:
@koaning All right, no problem :)

koaning:
I've played around with the tests locally and they make sense! Thanks! My only comment is that the tests feel kind of "big" in the sense that it is usually testing a lot of properties at a time. I don't think it's bad thought since the plotting functions themselves also support a lot of properties. 

There seems a *small* opportunity for refactoring but I prefer to consider that later after thinking seriously about it. For example; we might not need to test the width/height in every test. It feels verbose, but then again, it is no harm either. As is, these tests look legit to me so I'll gladly hit "merge". It also feels like we might refactor this code at some later point in time but this will likely be around the time when we overhaul the visualisation API. Better to wait for any serious refactoring 'till then.

Before I hit merge though, @mkaze is there a particular reason why you chose to test all these properties in allmost every test? Purely out of curiousity.

mkaze:
@koaning Yeah, what you said is totally correct. I have taken an extreme approach in testing the plot, either for altair or matplotlib plots, which may not be necessarily useful or beneficial compare to the case where you might have one single test case per each property of the plot (e.g. one for labels, one for title, one for each axis type, etc.).

Honestly, I was aware of this and even considered refactoring to some extent; however, the thing that somehow motivated me in the first place to take this approach was mostly the dependency of arguments' values and the plot properties. For example, by default the axis arguments determine the value of axis labels (and we have three axis kinds: integer, str, embedding), but setting `x_label` or `y_label` changes this. Of course, one might argue here that we can test all of these in a single test case that only concerns testing axis labels; and even further, there might be no need to test all combinations!

So, all in all, I agree that the tests could have been written using a more relaxed and simpler approach without end up being too verbose. And I would be willing to re-write or refactor them as such.

koaning:
Let's do the refactoring later, when we get 'round to the new visualisation API. Will merge this for now. 

Thanks! Having unit tests for these methods definately makes me sleep better at night 😅

PR #225, opened 2020-09-16
Fix #207 - Add Warnings for Axis Metrics

### LINKED ISSUE #207: Warning for axis_metric
Let's set up a small embeddingset. 

```python
from whatlies.language import SpacyLanguage
from whatlies.transformers import Pca

words = ["prince", "princess", "nurse", "doctor", "banker", "man", "woman",
         "cousin", "neice", "king", "queen", "dude", "guy", "gal", "fire",
         "dog", "cat", "mouse", "red", "bluee", "green", "yellow", "water",
         "person", "family", "brother", "sister"]

lang = SpacyLanguage("en_core_web_md")
emb = lang[words]
```

Let's now make some charts. 

```python
emb.transform(Pca(2)).plot(kind='scatter', annot=True)
emb.transform(Pca(2)).plot(kind='scatter', annot=True, axis_metric='cosine')
emb.transform(Pca(2)).plot(kind='scatter', annot=True, axis_metric='euclidean')
```

At the moment, they all seem to return this chart.

image

It took me a while to realise that the reason why this was not working as expected was because of the fact that I'm not using `"king"` as an x_axis/y_axis. Maybe we should introduce a warning here. Might be more user friendly.

koaning:
This PR fixes https://github.com/RasaHQ/whatlies/issues/207. 

I've added warnings to `EmbeddingSet.plot`, `EmbeddingSet.plot_3d`, `EmbeddingSet.plot_interactive` and `Embedding.plot`. When a user picks a `axis_metric` without specifying an embedding in a plot call they'll now get a warning that it is ignored. 

I've also added tests for all these.

### DIFF
--- a/tests/test_plotting/test_matplotlib/common.py
+++ b/tests/test_plotting/test_matplotlib/common.py
@@ -2,4 +2,5 @@ def validate_plot_general_properties(ax, props):
     assert ax.xaxis.get_label_text() == props["x_label"]
     assert ax.yaxis.get_label_text() == props["y_label"]
     assert ax.get_title() == props["title"]
+    print(ax.get_aspect())
     assert ax.get_aspect() == props["aspect"]
--- a/tests/test_plotting/test_matplotlib/common.py
+++ b/tests/test_plotting/test_matplotlib/common.py
@@ -2,5 +2,4 @@ def validate_plot_general_properties(ax, props):
     assert ax.xaxis.get_label_text() == props["x_label"]
     assert ax.yaxis.get_label_text() == props["y_label"]
     assert ax.get_title() == props["title"]
-    print(ax.get_aspect())
     assert ax.get_aspect() == props["aspect"]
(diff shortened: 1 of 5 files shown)


mkaze (commented):
Good! But should we also raise a warning if only one of the axes is an integer, like:
```python
embset.plot(x_axis='foo', axis_metric='euclidean')
```
Or:
```python
# euclidean metric would NOT be used for x-axis.
embset.plot(y_axis='foo', axis_metric=['euclidean', 'cosine_distance'])

# maybe the user is aware of this because of passing `None`?!
embset.plot(y_axis='foo', axis_metric=[None, 'cosine_distance']) 
```

koaning:
That's a good one. I think if the user is explicitly passing `None` that it is safe to say that the user is aware of the intention and that a warning should be avoided. There should be a way to turn off the warning and passing the explicit `None` here seems like a good way to do that. 

I'll add the warnings for the other two cases though 👍

Issue #204, opened 2020-09-08
Transformer API design and implementation


mkaze:
It's good to have a dedicated place for discussing ideas about the transformer API design and also the potential improvement/refactoring options in its implementation. Specifically, we are interested to find out what the drawbacks of the current API design (and/or implementation) is, and how it could be further improved.

**Note that if we an issue/idea is big enough to merit having a dedicated issue or we decide to implement a particular idea/feature, feel free to create a new specific issue for that to be able to track them easily.**

sara-tagger:
Thanks for the issue, @b-quachtran will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

sara-tagger:
Thanks for raising this issue, @TyDunn will get back to you about it soon✨

###### Please also check out the [docs](https://rasa.com/docs/) and the [forum](https://forum.rasa.com/) in case your issue was raised there too 🤗

sara-tagger:
Thanks for the issue, @rctatman will get back to you about it soon!

###### You may find help in the [docs](https://rasa.com/docs/) and the [forum,](https://forum.rasa.com/) too 🤗

koaning:
I'm currently working on a PR to help this disucssion. I might have found a nice balance API wise and I it might help the discussion to have a code example to point to.

koaning:
Good to mention: not every transformer will fit this API though. The `Noise` transformer seems like the exception. I've ignored `AddRandom` because I'm assuming that we're deprecating it (do we agree?).

koaning:
I'm closing issues because ever since the project moved to my personal account it's been more into maintenance mode than a "active work" mode.

PR #226, opened 2020-09-16
[WIP] #204 Transformer Abstraction

koaning:
The primary goal with this PR is to help the discussion in https://github.com/RasaHQ/whatlies/issues/204. If folks agree we might expand the idea here to a code-change.


mkaze (commented):
I like the idea and understand the abstraction.
However, one thing which I am not sure about is whether the definition of transformer itself should be as a function or a class. Since we are implementing this with an object-oriented design, defining them as a class makes more sense to me. For example:
```python
class Pca(SklearnTransformer):
    def __init__(self, n_components, **kwargs):
        super().__init__(PCA, "pca", n_components=n_components, *args, **kwargs)
```
This also open up the room for more transformer-specific modifications/additions in future.

koaning:
@mkaze you're right. Object oriented would be better. 

I'll go ahead and iterate using an object variant for `Pca`. Once we both agree that it's what we want I'll continue implementing the other ones.

whatlies/transformers/__init__.py
```diff
@@ -1,12 +1,29 @@
...
+
+
+def Pca(n_components=2, *args, **kwargs):
+    """
+    Creates a PCA transformer component that you can use on an `EmbeddingSet`.
+
+    You can pass it all the parameters from the underlying scikit-learn implementation for customisation.
+    """
+    return SklearnTransformer(
+        PCA, name="pca", n_components=n_components, *args, **kwargs
```
mkaze, line 23: The `name` should be passed as positional, not keyword argument here. Otherwise, the `*args` could not be used at all and becomes useless. For example, with current implementation, the following would raise an error:
```python
Pca(2, 0, 1, key='dummy')
```

koaning: This is making me wonder. Do we *really* want to allow for args here? There's something to be said to force the users to only pass arguments explicitly.

mkaze: I am also in favor of having only the `**kwargs`, since the positional arguments are usually important and therefore should be explicitly listed in the signature. In my opinion, the (only and) most useful case for having both is when the function is defined as `def func(*args, **kwargs): ...` with no additional arguments.

koaning: Agreed! I also think on the scikit-learn side of things they are becomming more strict when it comes to arguments. I recall version 0.24.0 onwards the API demands key-word arguments for gridsearch.

koaning:
There's an issue that seems related to [tqdm](https://github.com/tqdm/tqdm) which is a progress bar used by ivis. It also seems like this project just got an update according to [pypi](https://pypi.org/project/tqdm/#history).

koaning:
@mkaze no rush, but I think the branch is looking green on my end.

whatlies/transformers/_normalizer.py
```diff
@@ -1,6 +1,6 @@
 from sklearn.preprocessing import normalize
 
-from whatlies.transformers import Transformer
+from ._transformer import Transformer
 from whatlies import EmbeddingSet
 from whatlies.transformers._common import new_embedding_dict
```
mkaze, line 6: Same as `Noise`, it could have been refactored like this:
```python
    super().__init__(
        FunctionTransformer, f"normalizer_{norm}", func=lambda X: normalize(X, norm=norm, axis=0 if feature else 1), **kwargs
    )
```

whatlies/transformers/_opentsne.py
```diff
@@ -1,12 +1,12 @@
 import numpy as np
 from openTSNE import TSNE
 
-from whatlies.transformers import Transformer
+from ._transformer import SklearnTransformer
 from whatlies import EmbeddingSet
 from whatlies.transformers._common import new_embedding_dict
```
mkaze, line 6: Unused imports.

whatlies/transformers/_tsne.py
```diff
@@ -51,6 +50,8 @@ def fit(self, embset):
 
     def transform(self, embset):
         names, X = embset.to_names_X()
+        # We are re-writing the transform method here because TSNE cannot .fit().transform().
+        # Check the docs here: https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html#sklearn.manifold.TSNE
         new_vecs = self.tfm.fit_transform(X)
         new_dict = new_embedding_dict(names, new_vecs, embset)
-        return EmbeddingSet(new_dict, name=f"{embset.name}.tsne_{self.n_components}()")
+        return EmbeddingSet(new_dict, name=f"{embset.name}.tsne()")
```
mkaze, line 57: Why removing the `n_components` from the name, unlike other transformers? Maybe you have meant writing `f"{embset.name}.{self.name}()"` instead?!

whatlies/transformers/_tsne.py
```diff
@@ -38,10 +38,9 @@ class Tsne(Transformer):
...
     def __init__(self, n_components=2, **kwargs):
-        super().__init__()
-        self.n_components = n_components
-        self.kwargs = kwargs
-        self.tfm = TSNE(n_components=n_components, **kwargs)
+        super().__init__(
+            TSNE, f"tsne_{n_components}", n_components=n_components, **kwargs
+        )
 
     def fit(self, embset):
```
mkaze, line 45: This method redundant since it's a subclass of `SklearnTransformer` now.

mkaze (approved):
Thanks a lot! It looks good to me.
I just made some comments wherever I thought there is still some room for improvement, or I was curious to know the reason for the change.

whatlies/transformers/_transformer.py
```diff
@@ -37,3 +38,31 @@ def transform(self, embset: EmbeddingSet) -> EmbeddingSet:
...
+    def __init__(self, tfm, name, *args, **kwargs):
+        super().__init__()
+        self.tfm = tfm(*args, **kwargs)
+        self.name = name
+
+    def fit(self, embset: EmbeddingSet) -> "SklearnTransformer":
+        if not self.is_fitted:
+            # This is a bit of an anti-pattern. You should not need to `self.tfm`. However, there are
+            # some packages like OpenTSNE that return a different kind of estimator once an estimator has been fitted.
+            self.tfm = self.tfm.fit(embset.to_X())
```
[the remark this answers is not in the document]
koaning: If the transformer isn't fitted, it cannot `self.tfm.transform()`. It needs to be fitted at least once. By adding the `if not self.is_fitted` we prevent that we rerun the `self.tfm.fit`-command unless it is explicitly fitted again. 

The use-case here is that you might want to call `Pca` on a subset of data (say, heavily gendered words) to then use it again on the entire dataset.

whatlies/transformers/__init__.py
```diff
@@ -18,7 +21,6 @@
     Ivis = NotInstalled("Ivis", "ivis")
 
 __all__ = [
-    "Transformer",
```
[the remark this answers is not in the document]
koaning: Fair!

whatlies/transformers/_transformer.py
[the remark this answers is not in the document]
mkaze: @koaning All right, as you mentioned, that's the responsibility of `transform` method to check for this and first fit a not-yet-fitted transformer (which is already done in `transform` method). So I still don't understand the point of that check in `fit` method. In other words, if `fit` method is called **directly** it should fit the transformer on the given data, no matter it's been already fitted or not.

whatlies/transformers/_normalizer.py
```diff
@@ -1,18 +1,18 @@
...
+from sklearn.preprocessing import normalize, FunctionTransformer
 
-from whatlies.transformers import Transformer
+from ._transformer import SklearnTransformer
 from whatlies import EmbeddingSet
 from whatlies.transformers._common import new_embedding_dict
 
 
-class Normalizer(Transformer):
+class Normalizer(SklearnTransformer):
```
mkaze, line 8: @koaning Was there any problem, since I see that you have chosen to not remove the `fit` and `transform` methods after subclassing `SklearnTransformer`?

koaning: Well spotted! I'll check. I recall correcting that.

PR #229, opened 2020-09-16
Minor Feature: Add Negation Operator to Embedding

koaning:
I wanted to do:

```python
lang.score_similar(-lang['blue'])
```

But this caused an error. I figured adding `__neg__` is helpful. Allows for more "playfullness".

### DIFF
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -85,4 +85,4 @@ def test_emb_ndim():
 
 def test_negation():
     foo = Embedding("foo", [0.1, 0.3])
-    assert np.all(np.isclose((-foo).vector, -np.array([0.1, 0.3])))
+    assert np.allclose((-foo).vector, -np.array([0.1, 0.3]))
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -81,3 +81,8 @@ def test_emb_str_method(emb):
 def test_emb_ndim():
     foo = Embedding("foo", [0, 1, 0.2])
     assert foo.ndim == 3
+
+
+def test_negation():
+    foo = Embedding("foo", [0.1, 0.3])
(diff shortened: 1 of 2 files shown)


tests/test_embedding.py
```diff
@@ -81,3 +81,8 @@ def test_emb_str_method(emb):
 def test_emb_ndim():
     foo = Embedding("foo", [0, 1, 0.2])
     assert foo.ndim == 3
+
+
+def test_negation():
+    foo = Embedding("foo", [0.1, 0.3])
+    assert np.all(np.isclose((-foo).vector, -np.array([0.1, 0.3])))
```
mkaze, line 88: Just a reminder/shortcut: there is also `np.allclose`.

whatlies/embedding.py
```diff
@@ -93,6 +93,25 @@ def __sub__(self, other):
         copied.vector = self.vector - other.vector
         return copied
 
+    def __neg__(self):
+        """
+        Negate a embedding.
```
mkaze, line 98: "an" instead of "a".

whatlies/embedding.py
```diff
@@ -93,6 +93,25 @@ def __sub__(self, other):
...
+        ```python
+        from whatlies.embedding import Embedding
+
+        foo = Embedding("foo", [0.1, 0.3])
+
+        assert (- foo).vector == - foo.vector
+        ```
+        """
+        copied = deepcopy(self)
+        copied.name = f"(- {self.name})"
```
mkaze, line 111: I am not sure if the space between the negation and the name should be there or not. In math writing, we don't put it but here the situation might be a bit different.

mkaze (approved):
It looks good to me. Thanks for adding this!

PR #232, opened 2020-09-21
Minor sklearn docs patch

koaning:
This is a temorary branch to check if the CLA works as expected.


koaning:
The CLA seems to work. Closing.

PR #236, opened 2020-09-23
Skip Ivis tests temporarily

mkaze:
This PR temporarily comments out the Ivis tests until the external problem, which is probably due to `tqdm`, is resolved.


koaning:
Ah. I saw this after I merged the other one. My bad. Will close this one.

PR #240, opened 2020-10-07
Convert Deprecation

koaning:
ConveRT unfortunately got yanked. So we need to update the library. 

I've replaced it with the universal sentence encoder for now.


koaning:
Only a subset of universal sentence encoder models follow the API we have in the `TFHubLanguage`. I've added those because it was simple to do so. If demand increases we can consider adding the other ones but we need to consider adding support for google's sentence tokenizer.

koaning:
Adding this for now as a replacement for ConveRT.

Issue #278, opened 2021-02-02
Remove Transformers Module


koaning:
Nobody uses it and it's another dependency that can get removed.

koaning:
Dependencies are getting to me and when I originally designed the API I thought we'd have transformers that might add embeddings. I no longer think this is true and we basically only use embeddings that are sklearn compatible. I've decided to relay the responsibility to the user. 

Related: https://github.com/RasaHQ/whatlies/issues/276

koaning:
Removing everything would be too drastic. Keeping `Umap` around but turning it into an optional dependency.

Issue #276, opened 2021-01-24
Error message when trying to use PCA


talamoig:
Hi,
I've just discovered Whatlies library and I was able to produce some interesting visualization. Though as soon as I try to import the PCA transformers (`from whatlies.transformers import Pca`), I get an error. 
It also happens if it's the first line in the code, so I guess it must be somehow related to my specific modules/python combination.
I am running python 3.9.1 and here is the list of pip packages (I'm showing them all since I'm not sure about dependencies). Error message is afterwards.

Thanks,
Ivano

```
Package               Version
--------------------- -----------
altair                4.1.0
apparmor              3.0.1
appdirs               1.4.4
argon2-cffi           20.1.0
asn1crypto            1.4.0
async-generator       1.10
attrs                 20.3.0
backcall              0.2.0
bcc                   0.18.0
binaryornot           0.4.4
biplist               1.0.3
bleach                3.2.1
blis                  0.7.4
bob                   0.1
bpemb                 0.3.2
Brlapi                0.8.0
btrfsutil             5.9
CacheControl          0.12.6
catalogue             1.0.0
cffi                  1.14.4
chardet               3.0.4
click                 7.1.2
colorama              0.4.4
contextlib2           0.6.0.post1
cryptography          3.3.1
cycler                0.10.0
cymem                 2.0.5
decorator             4.4.2
defusedxml            0.6.0
distlib               0.3.1
distro                1.5.0
docutils              0.16
en-core-web-sm        2.3.1
entrypoints           0.3
fasttext              0.9.2
flashfocus            2.2.2
gensim                3.8.3
html5lib              1.1
i3ipc                 2.2.1
idna                  2.10
ipykernel             5.3.4
ipython               7.19.0
ipython-genutils      0.2.0
ipywidgets            7.5.1
it-core-news-lg       2.3.0
it-core-news-md       2.3.0
it-core-news-sm       2.3.0
jedi                  0.17.2
jeepney               0.6.0
Jinja2                2.11.2
joblib                0.17.0
jsonschema            3.2.0
jupyter               1.0.0
jupyter-client        6.1.7
jupyter-console       6.2.0
jupyter-core          4.7.0
jupyterlab-pygments   0.1.2
keyring               21.5.0
kiwisolver            1.3.1
LibAppArmor           3.0.1
libfdt                1.6.0
lit                   0.10.1.dev0
llvmlite              0.34.0
louis                 3.16.0
MarkupSafe            1.1.1
[...]

PR #280, opened 2021-02-16
Transformer Purge

koaning:
Let's make this package more lightweight. Some of the transformers are never ever used. And UMAP is causing errors.


koaning:
Fixes https://github.com/RasaHQ/whatlies/issues/278 and https://github.com/RasaHQ/whatlies/issues/276.

PR #298, opened 2021-03-22
add color ranked text in plot brush

RensDimmendaal:
Small suggestion to improve the readability of the brush plot.
Now you can als see to which group the ranked text belongs.
This is particularly helpful when your embedding+dimension reduction method isn't able to separate the different groups.

Current situation:
image

Proposal:
image

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -1318,7 +1318,10 @@ class EmbeddingSet:
         ranked_text = (
             alt.Chart(plot_df)
             .mark_text()
-            .encode(y=alt.Y("row_number:O", axis=None))
+            .encode(
+                y=alt.Y("row_number:O", axis=None),
+                color=alt.Color(":N", legend=None) if not color else alt.Color(color),
+            )
             .transform_window(row_number="row_number()")
             .transform_filter(brush)
             .transform_window(rank="rank(row_number)")


sara-tagger:
Thanks for submitting a pull request 🚀 @degiz will take a look at it as soon as possible ✨

koaning:
I like the suggestion! Let's discuss a few points. 

1. Would it make sense to order the points in the textbox? In your example the blue group is separate from the orange one, but does this depend on the order in the data frame? Since we only show `n` examples at a time, do we want to stratify the examples shown? Would it perhaps make sense to render a list per group instead? 
2. Can we turn this feature into a flag? I can imagine that situations with many groups that you would like to see the cluster in the chart but not in the text-box. 
3. Could you sign the CLA? Without it, I cannot merge anything into a Rasa repository 😅.

RensDimmendaal:
Thanks for the feedback

1. I suggest ordering them same as the y-axis of the scatter plot; that consistency makes it easier to identify which text belongs to which dot.
2. As in an additional flag to the color in the scatter plot? Fine by me of course, but as a user I'd be fine with them being toggled by the same flag.
3. Let me align with the day-job people to avoid future hassles.

RensDimmendaal:
fyi the approval takes a bit longer due to holidays. haven't forgotten about this. apologies for the delay.

koaning:
> As in an additional flag to the color in the scatter plot? Fine by me of course, but as a user I'd be fine with them being toggled by the same flag.

Yeah, you're right. Putting that in the same flag seems fine.

RensDimmendaal:
sorry for the hold up.

PR #302, opened 2021-04-02
Add pan and zoom functionality to plot_brush

nth-attempt:
Panning and Zooming seems useful even for selection kind of plots especially when we have a lot of dense points to select from.

Adding this functionality (when shift key is held) to `plot_brush` method

### DIFF
--- a/whatlies/embeddingset.py
+++ b/whatlies/embeddingset.py
@@ -1204,6 +1204,7 @@ class EmbeddingSet:
         annot: bool = False,
         color: Union[None, str] = None,
         n_show: int = 15,
+        interactive: bool = True,
     ):
         """
         Makes an interactive plot with a brush element.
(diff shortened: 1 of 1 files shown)


koaning:
I was completely unaware that you're able to add interactivity via keyboard support. I'll look into it on Tuesday (that's when I get back to work).

nth-attempt:
Here is the vega documentation for the same https://vega.github.io/vega/docs/event-streams/#selector

koaning:
The experience is a bit trippy, but all in all, I agree that for the select few who need to be able to zoom in here it should certainly suffice. Nicely done! 

I'll merge this PR right away but the release might take another week or two. I'd like to resolve the PR over here https://github.com/RasaHQ/whatlies/pull/298 before I push a new release.

PR #308, opened 2021-05-17
Update setup.py

koaning:


koaning:
It seems TensorFlow 2.5 is the culprit.

